PR #6, opened 2023-03-02
WebApi

HoLuDyEr:
uses Bored Api to get activities, has a cache to use it offline


damiengenoud:
Not needed (at least for now)

PR #33, opened 2023-03-07
15 add basic main activity

### LINKED ISSUE #15: Create a main activity with different "pages" like main/map, settings, profile and other

damiengenoud:
Adds the main activity skeleton and the main fragments (all empty, to be completed in the future), with two different types of menus : one at the bottom of the screen and one "drawer menu".

The goal is to have a design somewhat close to modern successful apps (like Instagram or even Reddit), but with views containing what our app might need in the future.  

Of course, all of this could be changed in the next weeks if some parts of this skeleton happened to not be required, so it should be seen as a way to start linking parts of the app rather than a fixed design. 

Anyone could review this code.

closes #15


clementcharmillot (changes requested):
Good PR, the main menu is well modularized and prepares well for the upcoming features.

Just two small remarks:
- I think it would be better to move the main menu fragments as well as the activity to its own subpackage called `mainmenu` or something similar.
- For the interface, the settings should probably go in the upper navigation bar as 5 items in the bottom navigation makes it quite crowded and it shouldn't be accessed that often. Putting it at the bottom of the left panel that appears after clicking the profile picture is also an option.

damiengenoud:
@clementcharmillot I agree with you on the package organization so I will change that, however I think that removing one item from the bottom menu is not "elegant" as the main page (map drawing) would not be in the center anymore with four items.
Also, an app like WhatsApp on IOS has the settings button placed exactly like in this design. Maybe we could get rid of the left drawer menu and put everything in the bottom menu on the right...
All of this should be discussed during a meeting but for now I think having a basis is important, and changes can be done very easily so I would not worry too much about that.

clementcharmillot (approved):
Thank you for the changes with the package.

Yes, removing the drawer and combining the submenus is also an option. Let's keep it like that now has it satisfies the requirements of the review, and discuss about it in a meeting.

PR #34, opened 2023-03-08
Google maps

### LINKED ISSUE #17: Finish team bootcamp Google Maps

Gafragni:
This is the basic configuration for all google maps implementations.
It took more time than expected due to issues with the API key and its encryption.

closes #17


damiengenoud (approved):
Looks good to me, don't forget to close any issue/task that this PR would complete.

app/src/main/AndroidManifest.xml
```diff
@@ -11,9 +14,16 @@
         android:supportsRtl="true"
         android:theme="@style/Theme.Bootcamp"
         tools:targetApi="31">
+        <meta-data
+            android:name="com.google.android.gms.version"
+            android:value="@integer/google_play_services_version" />
+        <meta-data
+            android:name="com.google.android.geo.API_KEY"
+            android:value="${MAPS_API_KEY}" />
```
damiengenoud, line 22: This line (22) creates an error on my end, because MAPS_API_KEY has no value. Are you able to launch the app ? If so, where does this value come from ?

Gafragni: The problem is that this variable is initialized in local.properties but this file is in the gitignore (so not on git).
The other problem is that we are not supposed to put the API key on github (which is why we have to do all those extra steps).

Others have had a similar problem on discord and from what I understand we are supposed to only have one key per project but not put it on git, so the only solution I can see is that I share the API key privately and ask everyone in the group to modify their local.properties with the API key.

The modification in local.properties is trivial, simply append MAPS_API_KEY= and the key value.

damiengenoud: Alright, send us the key value in our group chat with the modification to be made and then you can merge.

damiengenoud (approved):
Looks good to me.

Issue #23, opened 2023-03-03
Create an activity with a form so that a user can enter specific sport data and social data and this information will be directly store on the database.


HugoHof:
This form contains :  
- an activité that welcome the user and ask him to create a a user profile
- a fragment to ask the user to choose a username(that is not already taken by another user in the database) and test if the username is available with feedback to the user.
- a fragment to ask the user his first name, username and date(with a date picker) (each input is checked to verify the format of the date and the age condition 10<age<100)
- a fragment to indicate some daily goal: distance, activity time and number of paths(again the input are checked)
- a fragment to indicate to the user that his profile has been successfully created that contains the username chosen by the user.

PR #36, opened 2023-03-10
14 google signin

### LINKED ISSUE #14: Add sign in login with the Firebase backend for Google authentication to the login page

clementcharmillot:
This PR adds the following features to the login activity:

- One-Tap sign-in with Google on app startup
- Automatic login from keychain from previous runs
- Register with Google Sign-In
- Login with Google Sign-In
- A new code interface named `Auth` to manage authentication

A few code samples to get the `uid` of the current user:

```kotlin
val uid = FirebaseAuth.getUser().getUid()
```

Or with test support:

```kotlin
import com.github.drawyourpath.bootcamp.authentication.Auth
import com.github.drawyourpath.bootcamp.authentication.FirebaseAuth
import com.github.drawyourpath.bootcamp.authentication.MockAuth

class MyActivity : AppCompatActivity() {

    private lateinit var auth: Auth
    
    override fun onCreate(savedInstanceState: Bundle?) {
        super.onCreate(savedInstanceState)

        // ...
       
        val useMockAuthProvider = intent.getBooleanExtra(USE_MOCK_AUTH_KEY, false)
        auth = when (useMockAuthProvider) {
            true  -> MockAuth(useMockAuthProvider)
            false -> FirebaseAuth()
        }
    }

    // Example use 
    fun foo() {
        val uid = auth.getUser().getUid()
    }
}
```

Some tests need #23 to check for redirection after registering.

Coverage is low as the `FirebaseAuth` class can't be tested directly.

closes #14


app/src/main/java/com/github/drawyourpath/bootcamp/authentication/FirebaseAuth.kt
```diff
@@ -0,0 +1,236 @@
...
+                callback(convertUser(res.user!!), null)
+            }
+            .addOnFailureListener { exc ->
+                callback(null, exc)
+            }
+    }
+
+    override fun registerWithGoogle(activity: Activity, callback: AuthCallback) {
+        loginWithGoogle(activity, callback)
+    }
```
CristianSafta, line 123: Why is registerWithGoogle the same as loginWithGoogle?

app/src/main/java/com/github/drawyourpath/bootcamp/authentication/FirebaseAuth.kt
```diff
@@ -0,0 +1,236 @@
...
+                }
+            }
+            .addOnFailureListener(activity) { e ->
+                Log.d("GSI", e.localizedMessage!!)
+            }
+    }
+
+    override fun signOut() {
+        auth.signOut()
+    }
```
CristianSafta, line 169: Maybe set the currCallback to null?

app/src/main/java/com/github/drawyourpath/bootcamp/authentication/FirebaseAuth.kt
```diff
@@ -0,0 +1,236 @@
...
+            consumeCurrCallback(null, e)
+        }
+    }
+
+    override fun onActivityResult(activity: Activity, requestCode: Int, resultCode: Int, data: Intent?) {
+        when (requestCode) {
+            REQ_GSI     -> onGoogleSignInResult(activity, data)
+            REQ_ONE_TAP -> onOneTapSignInResult(activity, data)
+        }
+    }
```
CristianSafta, line 223: Maybe add error handling in case requestCode is not what we expect.

CristianSafta (approved):
Overall looks good and well structured. Just a few minor comments

app/src/main/java/com/github/drawyourpath/bootcamp/authentication/FirebaseAuth.kt
clementcharmillot: The Google Sign-In SDK doesn't make a difference between registering and signing in. Registering is just the first sign in.

app/src/main/java/com/github/drawyourpath/bootcamp/authentication/FirebaseAuth.kt
clementcharmillot: The method receives all the results of the activity so it might receive unrelated results with different request codes depending on the activity forwarding the results. Because of that, I don't think we should raise an error.

app/src/main/java/com/github/drawyourpath/bootcamp/authentication/FirebaseAuth.kt
clementcharmillot: `currCallback` is automatically cleared after the sign in procedure completes. If we set it to null in `signOut`, it will cause a sign in to not call its callback or just do nothing as it is already `null`.

PR #40, opened 2023-03-14
Issue#27

### LINKED ISSUE #27: Create a class that represents a run along with basic information about it

CristianSafta:
Added A class representing a run taken by a user.
The run includes information such as the path taken, distance, duration, timestamp, user, averageSpeed, and calorie burn.
Various methods are included to calculate and retrieve this information.

It provides basic functionality for now but provides a backbone for adding more functionality in the future as its needed

Closes #27


app/src/main/java/com/github/drawyourpath/bootcamp/path/Path.kt
HugoHof, line 1: You comment your function with "//Example of comment", the problem is that you can access to this comment when in function description when you will use it later. You can write your commenter like that:
/**
 *Example of comment
 */
private fun function()
You can use the "/**" shortcut on your keyboard to write this comments faster.

app/src/main/java/com/github/drawyourpath/bootcamp/path/Path.kt
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    //helper function to calculate distance between two points(in meters)
+    private fun distance(point1: LatLng, point2: LatLng): Double {
+        val lat1 = point1.latitude
+        val lat2 = point2.latitude
+        val lon1 = point1.longitude
+        val lon2 = point2.longitude
+        val R = 6371e3 // metres
+        val φ1 = lat1 * Math.PI / 180 // φ, λ in radians
```
HugoHof, line 49: Try to to not used mathematical characters like φ or λ for the code style implementation.

app/src/main/java/com/github/drawyourpath/bootcamp/path/Path.kt
```diff
@@ -0,0 +1,89 @@
...
+        val φ2 = lat2 * Math.PI / 180
+        val Δφ = (lat2 - lat1) * Math.PI / 180
+        val Δλ = (lon2 - lon1) * Math.PI / 180
+        val a = Math.sin(Δφ / 2) * Math.sin(Δφ / 2) +
+                Math.cos(φ1) * Math.cos(φ2) *
+                Math.sin(Δλ / 2) * Math.sin(Δλ / 2)
+        val c = 2 * Math.atan2(Math.sqrt(a), Math.sqrt(1 - a))
```
HugoHof, line 56: (same thing here, maybe write 2.0 instead of 2)

app/src/main/java/com/github/drawyourpath/bootcamp/path/Run.kt
```diff
@@ -0,0 +1,88 @@
...
+
+    //the distance of the run (in meters)
+    private var distance: Double = 0.0
+    //the duration of the run (in seconds)
+    private var duration: Long = 0L
+    //the average speed of the run (in meters per second)
+    private var averageSpeed: Double = 0.0
+    //the calorie burn of the run
+    private var calories: Int = 0
+
```
HugoHof, line 39: (Maybe you can add the time made by the runner in (minutes, seconds) to make 1 kilometer, this measure is used in the most of the running apps)

app/src/main/java/com/github/drawyourpath/bootcamp/path/Run.kt
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    private fun calculateDuration() {
+        duration = endTime - startTime
+    }
+
+
+    fun calculateCalorieBurn() {
+        calories = 0
+        //TODO implement later based on the characteristics of the user
```
HugoHof, line 54: Good idea to add a function to approximate the calorie later !

app/src/main/java/com/github/drawyourpath/bootcamp/path/Run.kt
HugoHof, line 1: Write you comment with /** like explain before, it will be helpful when we will reused this functions.

app/src/main/java/com/github/drawyourpath/bootcamp/path/Path.kt
HugoHof, line 1: You comment your function with "//Example of comment", the problem is that you can access to this comment when in function description when you will use it later. You can write your commenter like that:
/**
*Example of comment
*/
private fun function()
You can use the "/**" shortcut on your keyboard to write this comments faster.

build.gradle
HugoHof, line 1: You think that you shouldn't update this file since he was configure at the beginning of the project and I think it shouldn't change.

HugoHof (changes requested):
Your implementation will be very useful for the next of the implementation of the app. Your code is very structured and there are lots of comment, that's very good. You can change the comment format such that the comments can be seen in the function descriptions. That is a good idea to add some functions like  getPolyLine(). 
I'm remain at your disposal if you have any questions.

HugoHof (approved):
Thank you for the code improving

HugoHof (approved):

PR #41, opened 2023-03-14
26 protoype settings

### LINKED ISSUE #26: Implement a prototype settings page

damiengenoud:
This small PR simply adds the UI for the preferences/settings fragment, with no logic for now. There are two different fragment: the main preferences fragment (PreferencesFragment) and a fragment which can be used to modify the password (ModifyPasswordFragment). It also adds a way to disconnect the user.

For now this is a prototype of a settings page which can be modified easily.

Testing the preference fragment is tedious and "hacky", see https://stackoverflow.com/questions/45172505/testing-android-preferencefragment-with-espresso, and the accepted solution creates many duplicate classes that I have yet not been able to solve, thus there is no test for now.

PreferencesFragment screen:
Capture d’écran 2023-03-14 160438

ModifyPasswordFragment screen:
Capture d’écran 2023-03-14 160503

closes #26


clementcharmillot:
I like the look of the settings interface. Not a lot of actual code, but it still satisfies what we expected.
We can merge for me.

PR #42, opened 2023-03-15
18 challenge activity

### LINKED ISSUE #18: create a challenge activity with goals like distance, time, shape

HoLuDyEr:
In this PR, I created a new fragment linked to the mainPage. To access click on the userIcon then challenge

Everything is in a scroll view, which then are in a recyclerView.

It uses a class TemporaryUser which needs to be changed when we have a real user class.

image

First we have the DailyGoals, which can be changed by changeing the number on the right.
It will also create a new DailyGoal each day.

image

Then we have the Tournaments in which the user takes part in.
For now there is only a name, a description,  a start date and an en date

image

Finally the Trophies earned by the user.

closes #18


app/src/androidTest/java/com/epfl/drawyourpath/mainpage/fragments/ChallengeFragmentTest.kt
HugoHof, line 1: The tests aren't  commented. It's better to give a description of the tests on the top of them with the keyboard shortcut /** instead of giving detail inside the tests.

app/src/androidTest/java/com/epfl/drawyourpath/mainpage/fragments/ChallengeFragmentTest.kt
```diff
@@ -0,0 +1,393 @@
...
+        bundle.putSerializable("user", userOnlyGoal)
+
+        val scenario = FragmentScenario.launchInContainer(
+            ChallengeFragment::class.java,
+            bundle,
+            R.style.Theme_Bootcamp
+        )
+
+        //check that distance is correct
+        onView(withId(R.id.goals_view))
```
HugoHof, line 47: This three tests should be separated in three independent tests because they are related to different date. It will be useful when this will linked directly to the user account

app/src/androidTest/java/com/epfl/drawyourpath/mainpage/fragments/ChallengeFragmentTest.kt
```diff
@@ -0,0 +1,393 @@
...
+            .perform(RecyclerViewActions.scrollToPosition<RecyclerView.ViewHolder>(0))
```
HugoHof, line 81: Same comment here: the three tests should be separated

app/src/androidTest/java/com/epfl/drawyourpath/mainpage/fragments/ChallengeFragmentTest.kt
```diff
@@ -0,0 +1,393 @@
...
+        closeSoftKeyboard()
+        //check that the value is correctly changed
+        onView(withId(R.id.goals_view))
+            .perform(RecyclerViewActions.scrollToPosition<RecyclerView.ViewHolder>(0))
+            .check(matches(hasDescendant(withText("1.56/20.0"))))
+            .check(matches(hasDescendant(withText("20"))))
+            .check(matches(hasDescendant(withText("kilometers"))))
+
+        //change value of distance
+        onView(withId(R.id.goals_view))
```
HugoHof, line 189: I think that this test correspond to the time instead of the distance

app/src/main/AndroidManifest.xml
HugoHof, line 1: I think you should delete this blank line to evict some merge conflicts

app/src/main/java/com/epfl/drawyourpath/challenge/DailyGoal.kt
```diff
@@ -0,0 +1,106 @@
...
+     * create a new Empty DailyGoal based on the current one
+     *
+     * @return a new DailyGoal
+     */
+    fun createNewGoalFromThis(): DailyGoal {
+        return DailyGoal(this.distanceInKilometerGoal, this.timeInMinutesGoal, this.nbOfShapesGoal)
+    }
+
+    /**
+     * should only be used inside [DailyGoalViewAdapter]
```
HugoHof, line 28: you should add the correspondance between each value in the comments : 0-> distance , 1 -> activity time

HugoHof: or you can create an enumeration like explain before

app/src/main/java/com/epfl/drawyourpath/challenge/TemporaryUser.kt
```diff
@@ -0,0 +1,111 @@
...
+            return dailyGoals.first
+        }
+        val newDailyGoal = dailyGoals.first.createNewGoalFromThis()
+        dailyGoals.addFirst(newDailyGoal)
+        return newDailyGoal
+    }
+
+    /**
+     * function to call when a trophy is earned
+     * it will add the trophy along with the current date
```
HugoHof, line 41: Maybe you can considerate the case when the user win the trophy more than one time. For example, a user can win the same trophy each time he made a run of 10 km(the most of the running do that for simplification)

app/src/main/java/com/epfl/drawyourpath/challenge/Tournament.kt
```diff
@@ -0,0 +1,68 @@
...
+import java.time.LocalDateTime
+
+/**
+ * class representing tournaments that the user can take part in
+ */
+data class Tournament(
+    val name: String,
+    val description: String,
+    val startDate: LocalDateTime,
+    val endDate: LocalDateTime
```
HugoHof, line 14: Maybe add a list in class parameter to add all the participant

app/src/main/java/com/epfl/drawyourpath/challenge/TournamentViewAdapter.kt
```diff
@@ -0,0 +1,52 @@
...
+    RecyclerView.Adapter<TournamentViewAdapter.ViewHolder>() {
+
+    /**
+     * Custom view holder using a custom layout for tournaments
+     */
+    class ViewHolder(view: View) : RecyclerView.ViewHolder(view) {
+        val name: TextView
+        val description: TextView
+        val date: TextView
+
```
HugoHof, line 24: Maybe add a list for the tournament member

app/src/main/java/com/epfl/drawyourpath/challenge/Trophy.kt
```diff
@@ -0,0 +1,19 @@
...
+    val trophyName: String,
+    val description: String,
+    val imagePath: String = "trophy/trophy.png"
+) : java.io.Serializable {
+
+    THEFIRSTKM("The First Kilometer", "run a total of 1 kilometer"),
+    TENKM("10 Kilometers", "run a total of 10 kilometers"),
+    MARATHON("MARATHON", "run a total of 42.195 kilometer"),
+    THEFIRSTHOUR("The first Hour", "run for a total of 1 hour"),
+    THEFIRSTPATH("The first Path", "draw your fisrt path")
```
HugoHof, line 16: Maybe you can add a trophy for the Ralph marathon and the 5km who are very popular distance which are present in the most of the running competition.

app/src/main/java/com/epfl/drawyourpath/challenge/TrophyViewAdapter.kt
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * Custom view holder using a custom layout for trophies
+     */
+    class ViewHolder(view: View) : RecyclerView.ViewHolder(view) {
+        val text: TextView
+        val dateText: TextView
+        val image: ImageView
+        val assets: AssetManager
+
```
HugoHof, line 29: You can consider the number of time the user win the same trophy(like explain before)(the most of the running app used this concept)

app/src/main/res/layout/display_list_daily_goal.xml
```diff
@@ -0,0 +1,67 @@
...
+            android:id="@+id/goal_progress_text"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_centerInParent="true"
+            android:background="@android:color/transparent" />
+
+    </RelativeLayout>
+
+    <Space
+        android:layout_width="0dp"
```
HugoHof, line 37: Good idea to use spaces !

HugoHof (changes requested):

HugoHof (changes requested):
This fragment and the different are presented in Avery beautiful  fragment. The ode is very modular, it will be very helpful for the next step of the application process. I have made some advice on the implementation and the code. 
Tank you for your work. 
I stay at your attention if you have any question .

app/src/main/java/com/epfl/drawyourpath/challenge/DailyGoal.kt
```diff
@@ -0,0 +1,106 @@
...
+     * update the goal associated to pos with a new value
+     *
+     * @param value the value
+     * @param pos the position of the goal from 0 to [count] - 1
+     */
+    fun updateGoal(value: String, pos: Int) {
+        val doubleValue = value.toDoubleOrNull()
+
+        if (doubleValue == null || doubleValue <= 0) {
+            return
```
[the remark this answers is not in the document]
HoLuDyEr: The user cannot enter negative value nor anything that is not a double it is more of a double check to avoid errors that may crash the app

app/src/androidTest/java/com/epfl/drawyourpath/mainpage/fragments/ChallengeFragmentTest.kt
HoLuDyEr: Thank you for the suggestion, I added comments

app/src/androidTest/java/com/epfl/drawyourpath/mainpage/fragments/ChallengeFragmentTest.kt
HoLuDyEr: I separated them into different tests

app/src/main/java/com/epfl/drawyourpath/challenge/DailyGoal.kt
HoLuDyEr: This logic should only be useful in the class DailyGoalViewAdapter so I moved it there.

app/src/main/java/com/epfl/drawyourpath/challenge/DailyGoal.kt
```diff
@@ -0,0 +1,106 @@
+package com.epfl.drawyourpath.challenge
+
+import java.time.LocalDate
+
```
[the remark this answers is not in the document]
HoLuDyEr: As I moved this logic out of DailyGoal it is not useful anymore as we should just access the fields normally

app/src/main/java/com/epfl/drawyourpath/challenge/DailyGoal.kt
```diff
@@ -0,0 +1,106 @@
...
+        }
+    }
+
+    /**
+     * should only be used inside [DailyGoalViewAdapter]
+     * update the goal associated to pos with a new value
+     *
+     * @param value the value
+     * @param pos the position of the goal from 0 to [count] - 1
+     */
```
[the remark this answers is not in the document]
HoLuDyEr: As the other comments, I moved this function outside of this class and it is only used by DailyGoalViewAdapter to update the corresponding field from the editText which returns a String.

app/src/main/java/com/epfl/drawyourpath/challenge/DailyGoal.kt
```diff
@@ -0,0 +1,106 @@
...
+            1 -> timeInMinutesGoal = doubleValue
+            2 -> nbOfShapesGoal = doubleValue.toInt()
+            else -> return
+        }
+    }
+
+    /**
+     * should only be used inside [DailyGoalViewAdapter]
+     * get the value of the associated to the goal
+     *
```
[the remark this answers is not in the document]
HoLuDyEr: Thank you for spotting it

app/src/main/java/com/epfl/drawyourpath/challenge/DailyGoal.kt
```diff
@@ -0,0 +1,106 @@
...
+            1 -> timeInMinutesGoal
+            2 -> nbOfShapesGoal.toDouble()
+            else -> 0.0
+        }
+    }
+
+    /**
+     * should only be used inside [DailyGoalViewAdapter]
+     * get the value of the associated to the progress towards the goal
+     *
```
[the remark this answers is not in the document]
HoLuDyEr: I changed it to 
"get the value of the current progress of the DailyGoal associated to pos"
I hope it is clearer

app/src/main/java/com/epfl/drawyourpath/challenge/Tournament.kt
HoLuDyEr: yes we need it but we do not have a user class yet so it should be added later.

app/src/main/java/com/epfl/drawyourpath/challenge/TournamentViewAdapter.kt
HoLuDyEr: It should be updated as soon as we have participants to display.

app/src/main/java/com/epfl/drawyourpath/challenge/Trophy.kt
HoLuDyEr: For now, those are placeholder, it will eventually be more populated

app/src/main/res/layout/display_list_daily_goal.xml
HoLuDyEr: Thanks

app/src/main/java/com/epfl/drawyourpath/challenge/TemporaryUser.kt
HoLuDyEr: I  do not know exactly how the trophy system will work but it is easy to change as we could change the LocalDate into a list of LocalDate but the placeholder I used were supposed to be a total distance over the time instead of a one time run.

app/src/androidTest/java/com/epfl/drawyourpath/mainpage/fragments/ChallengeFragmentTest.kt
HugoHof: Thanks for your change, the rest result will be more easy to understand.

app/src/main/java/com/epfl/drawyourpath/challenge/DailyGoalViewAdapter.kt
HugoHof, line 1: For the most of the method of this class, you never indicate the correspondance between the number 0-> distanceGoal, 1->time, 2->numberOfPaths. We cam^n guess it by reading the code, but it can be more easier if you create a get method or enum to returns this correspondances.

HugoHof (changes requested):
You take into account the most of my comments. Thank you for your time. There is just one thing to modify before merging. We can discuss it if you want-

app/src/main/java/com/epfl/drawyourpath/challenge/DailyGoal.kt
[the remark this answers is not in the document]
HugoHof: Later you will need to add a method which return a double to retrieve the data.

app/src/main/java/com/epfl/drawyourpath/challenge/DailyGoal.kt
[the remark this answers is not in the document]
HugoHof: So maybe throw an error instead of returning nothing

app/src/main/java/com/epfl/drawyourpath/challenge/TemporaryUser.kt
HugoHof: OK, I understand, We will discuss during the next meeting

app/src/main/java/com/epfl/drawyourpath/challenge/DailyGoalViewAdapter.kt
HoLuDyEr: I added an enumeration, it is indeed clearer now

HugoHof (approved):
Thank you for your change. It's more readable now. Thank you for your serious job

PR #43, opened 2023-03-16
25 past runs activity

### LINKED ISSUE #25: Create an activity that shows the user's saved paths and basic information about them

CristianSafta:
Modified the History fragment to display a list of past runs along with information about them
image

Along with the XML resource files, an adapter class was also created to display the information from the Run class into the History layout.

For now, a dummy list of runs is displayed until we have a way to store and retrieve the runs of each user.

Closes #25


clementcharmillot:
Just a small note that new code shouldn't be in the `bootcamp` folder.

PR #63, opened 2023-03-19
48 user model

HugoHof:
In this PR, you will fine a UserModel class that represent a user model for the app. This class contains: 
- a constructor to initialize a UserModel that will be use a the end of the profile creation(all the input are controlled with their format and dependencies with the database)
- get functions to return each parameter of a user
- set functions to edit the data Thant can be edited after the profile creation(daily goals, mail, username...) with controls on the new inputs
- getAge function to obtain the age of the user
- the structure of the friendliest of the user that will be implemented during task #49 and #58 

Remark: Some function will be modify of they are related to the debase in task #49 (I have added TODO commenter if it is the case). 

This task will be used for tasks: #49, #54, #55 (some of this task will need to wait the modification of the database #49 (see TODO comment if the method will be modify).

Tank you in advance for your review

### DIFF
--- a/app/src/main/java/com/epfl/drawyourpath/database/Database.kt
+++ b/app/src/main/java/com/epfl/drawyourpath/database/Database.kt
@@ -14,6 +14,13 @@ abstract class Database {
     abstract fun isUserStoredInDatabase(userId: String): CompletableFuture<Boolean>
 
     /**
+     * This function function will return a future that give the username in function of the userId
+     * @param userId of the user
+     * @return a future that give the usename of the user
+     */
+    abstract fun getUsernameFromUserId(userId: String): CompletableFuture<String>
+
+    /**
      * This function will return a future with a boolean to know if the username is available in the database
      * (i.e the userName proposed is not already associated to another user profile)
      * @param userName userName that the user want to use for his user profile
@@ -67,20 +74,23 @@ abstract class Database {
      * This function will set the daily distance goal to the database(in kilometer)
      * @param userId that correspond to the user
      * @param distanceGoal new distance goal of the user
+     * @return a future that indicate if the goal has been correctly set to the database
      */
-    abstract  fun setDistanceGoal(userId: String, distanceGoal: Double)
+    abstract  fun setDistanceGoal(userId: String, distanceGoal: Double): CompletableFuture<Boolean>
 
     /**
      * This function will set the daily activity time goal to the database(in minutes)
      * @param userId that correspond to the user
      * @param activityTimeGoal new activity time goal of the user
+     * @return a future that indicate if the goal has been correctly set to the database
      */
-    abstract  fun setActivityTimeGoalGoal(userId: String, activityTimeGoal: Double)
+    abstract  fun setActivityTimeGoalGoal(userId: String, activityTimeGoal: Double): CompletableFuture<Boolean>
 
     /**
      * This function will set the daily number of paths goal to the database(integer)
      * @param userId that correspond to the user
      * @param nbOfPathsGoal new number of paths goal of the user
+     * @return a future that indicate if the goal has been correctly set to the database
      */
-    abstract  fun setNbOfPathsGoalGoal(userId: String, nbOfPathsGoal: Double)
+    abstract  fun setNbOfPathsGoalGoal(userId: String, nbOfPathsGoal: Int):CompletableFuture<Boolean>
 }
--- a/app/src/main/java/com/epfl/drawyourpath/database/Database.kt
+++ b/app/src/main/java/com/epfl/drawyourpath/database/Database.kt
@@ -22,6 +22,14 @@ abstract class Database {
     abstract fun isUserNameAvailable(userName: String): CompletableFuture<Boolean>
 
     /**
+     * This function will update the username of the user if the username proposed is available(not taken by another user)
+     * @param username prposed by the user(will be set to the userprofile if it's available)
+     * @param userId userId of the user
+     * @return the future that indicate if the username have been updated
+     */
+    abstract fun updateUsername(username: String,userId: String): CompletableFuture<Boolean>
+
+    /**
      * This function will add the user Name to the database to create a new user profile
      * @param userName userName that the user want to set in the database
      */
--- a/app/src/main/java/com/epfl/drawyourpath/database/Database.kt
+++ b/app/src/main/java/com/epfl/drawyourpath/database/Database.kt
@@ -84,7 +84,7 @@ abstract class Database {
      * @param activityTimeGoal new activity time goal of the user
      * @return a future that indicate if the goal has been correctly set to the database
      */
-    abstract  fun setActivityTimeGoalGoal(userId: String, activityTimeGoal: Double): CompletableFuture<Boolean>
+    abstract  fun setActivityTimeGoal(userId: String, activityTimeGoal: Double): CompletableFuture<Boolean>
 
     /**
      * This function will set the daily number of paths goal to the database(integer)
@@ -92,5 +92,5 @@ abstract class Database {
      * @param nbOfPathsGoal new number of paths goal of the user
      * @return a future that indicate if the goal has been correctly set to the database
      */
-    abstract  fun setNbOfPathsGoalGoal(userId: String, nbOfPathsGoal: Int):CompletableFuture<Boolean>
+    abstract  fun setNbOfPathsGoal(userId: String, nbOfPathsGoal: Int):CompletableFuture<Boolean>
 }
--- a/app/src/main/java/com/epfl/drawyourpath/database/Database.kt
+++ b/app/src/main/java/com/epfl/drawyourpath/database/Database.kt
@@ -11,7 +11,7 @@ abstract class Database {
      * @param userId that correspond to the user
      * @return a future that indicate if the user is store on the database
      */
-    abstract fun isUserStoreOnDatabase(userId: String): CompletableFuture<Boolean>
+    abstract fun isUserStoredInDatabase(userId: String): CompletableFuture<Boolean>
 
     /**
      * This function will return a future with a boolean to know if the username is available in the database
@@ -19,7 +19,7 @@ abstract class Database {
      * @param userName userName that the user want to use for his user profile
      * @return the future that indicate if the username is available
      */
-    abstract fun isUserNameAvailable(userName: String): CompletableFuture<Boolean>
+    abstract fun isUsernameAvailable(userName: String): CompletableFuture<Boolean>
 
     /**
      * This function will update the username of the user if the username proposed is available(not taken by another user)
@@ -31,9 +31,9 @@ abstract class Database {
 
     /**
      * This function will add the user Name to the database to create a new user profile
-     * @param userName userName that the user want to set in the database
+     * @param username username that the user want to set in the database
      */
-    abstract fun setUserName(userName: String)
+    abstract fun setUsername(username: String)
 
     /**
      * This function will set the personal info of a user to the database (firstname, surname, date of birth).
@@ -58,8 +58,29 @@ abstract class Database {
      */
     abstract fun setUserGoals(
         username: String,
-        distanceGoal: Int,
-        timeGoal: Int,
+        distanceGoal: Double,
+        timeGoal: Double,
         nbOfPathsGoal: Int
     )
+
+    /**
+     * This function will set the daily distance goal to the database(in kilometer)
+     * @param userId that correspond to the user
+     * @param distanceGoal new distance goal of the user
+     */
+    abstract  fun setDistanceGoal(userId: String, distanceGoal: Double)
+
+    /**
+     * This function will set the daily activity time goal to the database(in minutes)
+     * @param userId that correspond to the user
+     * @param activityTimeGoal new activity time goal of the user
+     */
+    abstract  fun setActivityTimeGoalGoal(userId: String, activityTimeGoal: Double)
+
+    /**
+     * This function will set the daily number of paths goal to the database(integer)
+     * @param userId that correspond to the user
+     * @param nbOfPathsGoal new number of paths goal of the user
+     */
+    abstract  fun setNbOfPathsGoalGoal(userId: String, nbOfPathsGoal: Double)
 }
--- a/app/src/main/java/com/epfl/drawyourpath/database/Database.kt
+++ b/app/src/main/java/com/epfl/drawyourpath/database/Database.kt
@@ -7,6 +7,13 @@ import java.util.concurrent.CompletableFuture
 
 abstract class Database {
     /**
+     * This function is used to know if a certain user is already store in the database
+     * @param userId that correspond to the user
+     * @return a future that indicate if the user is store on the database
+     */
+    abstract fun isUserStoreOnDatabase(userId: String): CompletableFuture<Boolean>
+
+    /**
      * This function will return a future with a boolean to know if the username is available in the database
      * (i.e the userName proposed is not already associated to another user profile)
      * @param userName userName that the user want to use for his user profile
--- a/app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
+++ b/app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
@@ -10,7 +10,7 @@ private val TIMEOUT_SERVER_REQUEST: Long = 10
 
 class FireDatabase : Database() {
     val database: DatabaseReference = Firebase.database.reference
-    override fun isUserStoreOnDatabase(userId: String): CompletableFuture<Boolean> {
+    override fun isUserStoredInDatabase(userId: String): CompletableFuture<Boolean> {
         val future = CompletableFuture<Boolean>()
 
         database.child("users").child(userId).get().addOnSuccessListener {
@@ -22,7 +22,7 @@ class FireDatabase : Database() {
         return future
     }
 
-    override fun isUserNameAvailable(userName: String): CompletableFuture<Boolean> {
+    override fun isUsernameAvailable(userName: String): CompletableFuture<Boolean> {
         val future = CompletableFuture<Boolean>()
 
         database.child("users").child(userName).get().addOnSuccessListener {
@@ -38,9 +38,9 @@ class FireDatabase : Database() {
         TODO("This will be implmented during next task when I will clean the database organisation")
     }
 
-    override fun setUserName(userName: String) {
+    override fun setUsername(username: String) {
         val userAdd = HashMap<String, String>()
-        userAdd.put(userName, "empty")
+        userAdd.put(username, "empty")
         database.child("users").updateChildren(userAdd as Map<String, Any>)
             .addOnFailureListener { throw Exception("Impossible to add the username on the database") }
     }
@@ -65,8 +65,8 @@ class FireDatabase : Database() {
 
     override fun setUserGoals(
         username: String,
-        distanceGoal: Int,
-        timeGoal: Int,
+        distanceGoal: Double,
+        timeGoal: Double,
         nbOfPathsGoal: Int
     ) {
         val userAdd = HashMap<String, String>()
@@ -76,12 +76,25 @@ class FireDatabase : Database() {
         updateUserData(userAdd, username)
     }
 
+    override fun setDistanceGoal(userId: String, distanceGoal: Double) {
+        TODO("Will be done in the next task: database rebase")
+    }
+
+    override fun setActivityTimeGoalGoal(userId: String, activityTimeGoal: Double) {
+        TODO("Will be done in the next task: database rebase")
+    }
+
+    override fun setNbOfPathsGoalGoal(userId: String, nbOfPathsGoal: Double) {
+        TODO("Will be done in the next task: database rebase")
+    }
+
     /**
      * Helper functions to add some data to a user profile withe his username to the database
      * @param data date to add to the database
      * @param username corresponding to the user profile
      */
     private fun updateUserData(data: Map<String, Any>, username: String) {
+        //TODO:Fix this function in the nxt task such that the user profile is associated to the userId
         database.child("users").child(username).updateChildren(data)
             .addOnFailureListener {
                 throw Exception(buildString {
--- a/app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
+++ b/app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
@@ -84,11 +84,11 @@ class FireDatabase : Database() {
         TODO("Will be done in the next task: database rebase")
     }
 
-    override fun setActivityTimeGoalGoal(userId: String, activityTimeGoal: Double): CompletableFuture<Boolean> {
+    override fun setActivityTimeGoal(userId: String, activityTimeGoal: Double): CompletableFuture<Boolean> {
         TODO("Will be done in the next task: database rebase")
     }
 
-    override fun setNbOfPathsGoalGoal(userId: String, nbOfPathsGoal: Int): CompletableFuture<Boolean> {
+    override fun setNbOfPathsGoal(userId: String, nbOfPathsGoal: Int): CompletableFuture<Boolean> {
         TODO("Will be done in the next task: database rebase")
     }
 
--- a/app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
+++ b/app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
@@ -34,6 +34,10 @@ class FireDatabase : Database() {
         return future
     }
 
+    override fun updateUsername(username: String, userId: String): CompletableFuture<Boolean> {
+        TODO("This will be implmented during next task when I will clean the database organisation")
+    }
+
     override fun setUserName(userName: String) {
         val userAdd = HashMap<String, String>()
         userAdd.put(userName, "empty")
--- a/app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
+++ b/app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
@@ -10,6 +10,17 @@ private val TIMEOUT_SERVER_REQUEST: Long = 10
 
 class FireDatabase : Database() {
     val database: DatabaseReference = Firebase.database.reference
+    override fun isUserStoreOnDatabase(userId: String): CompletableFuture<Boolean> {
+        val future = CompletableFuture<Boolean>()
+
+        database.child("users").child(userId).get().addOnSuccessListener {
+            if (it.value == null) future.complete(true)
+            else future.complete(false)
+        }.addOnFailureListener {
+            future.completeExceptionally(it)
+        }
+        return future
+    }
 
     override fun isUserNameAvailable(userName: String): CompletableFuture<Boolean> {
         val future = CompletableFuture<Boolean>()
--- a/app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
+++ b/app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
@@ -22,6 +22,10 @@ class FireDatabase : Database() {
         return future
     }
 
+    override fun getUsernameFromUserId(userId: String): CompletableFuture<String> {
+        TODO("Will be implemented in the rebase database task")
+    }
+
     override fun isUsernameAvailable(userName: String): CompletableFuture<Boolean> {
         val future = CompletableFuture<Boolean>()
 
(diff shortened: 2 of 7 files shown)


app/src/main/java/com/epfl/drawyourpath/database/Database.kt
```diff
@@ -6,6 +6,13 @@ import java.util.*
 import java.util.concurrent.CompletableFuture
 
 abstract class Database {
+    /**
+     * This function is used to know if a certain user is already store in the database
+     * @param userId that correspond to the user
+     * @return a future that indicate if the user is store on the database
+     */
+    abstract fun isUserStoreOnDatabase(userId: String): CompletableFuture<Boolean>
```
clementcharmillot, line 14: I think there is a typo and it should be `isUserStoredInDatabase()`.

app/src/main/java/com/epfl/drawyourpath/database/Database.kt
```diff
@@ -14,6 +21,14 @@ abstract class Database {
      */
     abstract fun isUserNameAvailable(userName: String): CompletableFuture<Boolean>
```
clementcharmillot, line 22: Inconsistency in casing. Should be `isUsernameAvailable()` to match other function names.

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
```diff
@@ -10,6 +10,17 @@ private val TIMEOUT_SERVER_REQUEST: Long = 10
 
 class FireDatabase : Database() {
     val database: DatabaseReference = Firebase.database.reference
+    override fun isUserStoreOnDatabase(userId: String): CompletableFuture<Boolean> {
```
clementcharmillot, line 13: Should probably fix `private fun updateUserData(data: Map<String, Any>, username: String)` to use the `uid` instead and use this function here.

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
```diff
@@ -23,6 +34,10 @@ class FireDatabase : Database() {
         return future
     }
 
+    override fun updateUsername(username: String, userId: String): CompletableFuture<Boolean> {
+        TODO("This will be implmented during next task when I will clean the database organisation")
+    }
+
     override fun setUserName(userName: String) {
```
clementcharmillot, line 41: Inconsistency in casing. Should be `setUsername()` to match other function names

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
```diff
@@ -0,0 +1,354 @@
...
+     * @param firstname must respect the name convention of the app (name or name-name)
+     * @param surname must respect the name convention of the app (name or name-name)
+     * @param dateOfBirth the user be aged between 10 and 100 years old
+     * @param distanceGoal init at the user profile creation and can be modify after(daily goal)
+     * @param activityTimeGoal init at the user profile creation and can be modify after(daily goal)
+     * @param nbOfPathsGoal init at the user profile creation and can be modify after(daily goal)
+     * @throws an error if the inputs are incorrect
+     */
+    constructor(userId: String, username: String, emailAddress: String, firstname: String, surname: String, dateOfBirth: LocalDate, distanceGoal: Double, activityTimeGoal: Double, nbOfPathsGoal: Int, database: Database){
+        this.database = database
```
clementcharmillot, line 47: The class already has setters that handle invalid cases. We should call them instead of handling errors twice.

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
```diff
@@ -0,0 +1,354 @@
...
+    fun getDistanceGoal(): Double{
+        return distanceGoal
+    }
+
+    /**
+     * Use this function to modify the daily distance goal of the user
+     * @param distance new daily distance goal
+     */
+    fun setDistanceGoal(distance: Double){
+        if(distance == 0.0){
```
clementcharmillot, line 196: What if `distance < 0.0`?

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
```diff
@@ -0,0 +1,354 @@
...
+    fun getActivityTime(): Double{
+        return activityTimeGoal
+    }
+
+    /**
+     * Use this function to modify the daily activity time goal of the user
+     * @param time new daily activity time goal
+     */
+    fun setActivityTimeGoal(time: Double){
+        if(time == 0.0){
```
clementcharmillot, line 215: What if `time < 0.0`?

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
```diff
@@ -0,0 +1,354 @@
...
+private fun checkUsername(userId: String, username: String, database: Database): Boolean{
+    //TODO: Will be change in my next task when change the database organization
+    return !database.isUserNameAvailable(username).get(10, TimeUnit.SECONDS)
+}
+/**
+ * Helper function to check that the firstname and the surname respect the name format condition of the app
+ * @param name to be checked
+ * @return true if the name is in the correct format, and false otherwise
+ */
+private fun checkName(name: String): Boolean{
```
clementcharmillot, line 304: `checkName` is a bit vague, maybe rename it `checkNameFormat`.

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
```diff
@@ -0,0 +1,354 @@
...
+    }
+    return true
+}
+
+/**
+ * Helper function to check if the email address is correct
+ * @param email to be checked
+ * @return true is the email is in the correct format, and false otherwise
+ */
+private fun checkEMail(email: String):Boolean {
```
clementcharmillot, line 316: Inconsistency in casing, should be named `checkEmail()`.

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
```diff
@@ -0,0 +1,354 @@
...
+    // check for @ char
```
clementcharmillot, line 317: Going with our own email address checker is risky. Here, it typically accepts `mail@mail@mail.com`.
The doc shows the existence of `android.util.Patterns.EMAIL_ADDRESS.matcher(email).matches()` which might be more robust.

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
```diff
@@ -0,0 +1,354 @@
...
+
+    return true
+}
+/**
+ * Helper function to check the date of birth of the user
+ * @param date of the user birth
+ * @return return true if the user is aged between 10 and 100 years and false otherwise
+ */
+private fun checkDateOfBirth(date: LocalDate): Boolean{
+    if(date == null){
```
clementcharmillot, line 338: Could be replaced by 
```kotlin
return date !== null &&
    date < LocalDate.now().plusYears(-10) && 
    date > LocalDate.now().plusYears(-100)
```

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
```diff
@@ -0,0 +1,354 @@
...
+    fun getNumberOfPathsGoal(): Int{
+        return nbOfPathsGoal
+    }
+
+    /**
+     * Use this function to modify the daily number of paths goal of the user
+     * @param nbOfPaths new daily number of paths goal
+     */
+    fun setNumberOfPathsGoal(nbOfPaths: Int){
+        if(nbOfPaths == 0){
```
clementcharmillot, line 234: What if `nbOfPaths < 0`?

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
```diff
@@ -0,0 +1,354 @@
...
+     */
+    fun getEmailAddress(): String{
+        return emailAddress
+    }
+
+    /**
+     * Use this function to modify the email address of the user
+     * @param email new email address
+     */
+    fun setEmailAddress(email: String){
```
clementcharmillot, line 152: Email change should be forwarded to the authentication backend for account registered with emails. It should also be updated in the database for others.

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
```diff
@@ -0,0 +1,354 @@
...
+        return activityTimeGoal
+    }
+
+    /**
+     * Use this function to modify the daily activity time goal of the user
+     * @param time new daily activity time goal
+     */
+    fun setActivityTimeGoal(time: Double){
+        if(time == 0.0){
+            throw java.lang.Error("The activity time goal can't be equal to 0 !")
```
clementcharmillot, line 216: Should update the goal in the database.

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
```diff
@@ -0,0 +1,354 @@
...
+        return nbOfPathsGoal
+    }
+
+    /**
+     * Use this function to modify the daily number of paths goal of the user
+     * @param nbOfPaths new daily number of paths goal
+     */
+    fun setNumberOfPathsGoal(nbOfPaths: Int){
+        if(nbOfPaths == 0){
+            throw java.lang.Error("The number of paths goal can't be equal to 0 !")
```
clementcharmillot, line 235: Should update the goal in the database.

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
```diff
@@ -0,0 +1,354 @@
...
+        }
+        return age
+    }
+
+    /**
+     * This function will remove the user with username to the friend list
+     * @param username of the user that we want to remove
+     */
+    fun removeFriend(username: String) {
+        if(!friendsList.contains(username)) {
```
clementcharmillot, line 257: Should update the friends in the database. (Maybe add a TODO as it is in a following PR).

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
```diff
@@ -0,0 +1,354 @@
...
+    }
+
+}
+
+/**
+ * Helper function to check the userId and affect it if it's correct
+ * @param userId that correspond to the user
+ * @param database where the userId should be present
+ */
+private fun checkUserId(userId: String, database: Database):Boolean{
```
clementcharmillot, line 286: Doing a synchronous check is not possible here.

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
```diff
@@ -0,0 +1,354 @@
...
+    return database.isUserStoreOnDatabase(userId).get(10, TimeUnit.SECONDS)
+}
+/**
+ * Helper function to check the username and affect it if it's correct
+ * @param userId that correspond to the user
+ * @param username associated to the user profile
+ * @param database where the userId should be present
+ */
+private fun checkUsername(userId: String, username: String, database: Database): Boolean{
```
clementcharmillot, line 295: Doing a synchronous check is not possible here.

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
```diff
@@ -0,0 +1,354 @@
...
+
+import android.graphics.Color
+import android.text.TextUtils
+import com.epfl.drawyourpath.database.Database
+import java.time.LocalDate
+import java.util.concurrent.TimeUnit
+
+class UserModel {
+    //the userId is given by the authentication part and is unique to the user
+    private val userId: String
```
clementcharmillot, line 11: I think it's not a good idea to store the `uid` here. The `auth` object ensures a fresh `uid` that handle disconnecting or account change whenever we access it. I would suggest to simply forward the `uid` from the `Auth` instance.

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
```diff
@@ -0,0 +1,354 @@
...
+            throw java.lang.Error("The number of paths goal can't be equal to 0 !")
+        }
+        this.nbOfPathsGoal=nbOfPaths
+    }
+
+    /**
+     * Get the age of the user
+     * @return the age of the user
+     */
+    fun getAge(): Int{
```
clementcharmillot, line 244: I'd suggest to return a `Period` instead as it doesn't lose the month.

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
```diff
@@ -0,0 +1,354 @@
...
+     */
+    fun getUsername(): String{
+        return username
+    }
+
+    /**
+     * Use this function to modify the username(the username will be modify only if it is available on the database)
+     * @param username that we want to set
+     */
+    fun setUsername(username: String){
```
clementcharmillot, line 133: We might want to know when the operation completes to show a loading screen to the user, so we should probably return a `Future<Boolean>`. This is the same for all setters of this class.

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
```diff
@@ -0,0 +1,354 @@
...
+     * @param username is chosen during the profile create and each user has  unique username and it can be change later(if available in the database = not taken by another user)
+     * @param firstname must respect the name convention of the app (name or name-name)
+     * @param surname must respect the name convention of the app (name or name-name)
+     * @param dateOfBirth the user be aged between 10 and 100 years old
+     * @param distanceGoal init at the user profile creation and can be modify after(daily goal)
+     * @param activityTimeGoal init at the user profile creation and can be modify after(daily goal)
+     * @param nbOfPathsGoal init at the user profile creation and can be modify after(daily goal)
+     * @throws an error if the inputs are incorrect
+     */
+    constructor(userId: String, username: String, emailAddress: String, firstname: String, surname: String, dateOfBirth: LocalDate, distanceGoal: Double, activityTimeGoal: Double, nbOfPathsGoal: Int, database: Database){
```
clementcharmillot, line 46: As we have a lot of parameters, it might be worth it to create a builder class.

clementcharmillot (changes requested):
Good PR with good testing. I have a few remarks to address concerning the code.

I think the `UserModel` should follow the `write-through` model to avoid invalid data locally and take advantage of Realtime Database pricing method.

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
HugoHof: I will do that in my next task(rebase the database). I have added a TODO comment to think about it. Thanks

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
HugoHof: I have tested and the function matches doesn't work with this API (I have the same problem with the name format where I have try to used the matcher with REGEX)

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
HugoHof: I have commented this method since at this time we will need to modify some authentification class to implement the set of a new email.

HugoHof: We consider at this time that we can't modify the email address

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
clementcharmillot: I've tested 
```kotlin
private fun checkEmail(email: String):Boolean {
    return android.util.Patterns.EMAIL_ADDRESS.matcher(email).matches()
}
```
and it seems to be working as expected.

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
clementcharmillot: Actually, checking for `null` here is useless due to Kotlin's typing.

```kotlin
return date < LocalDate.now().plusYears(-10) && date > LocalDate.now().plusYears(-100)
```

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
HugoHof: OK, thank you. It's modified.

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
HugoHof: Thank you for the advice, it's modified

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
HugoHof: I think that return a Int is good because only the age in years in useful for the app.

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
HugoHof: I have change it with a return of a CompletableFure<Boolean>

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
HugoHof: I have delete it because it is useless since we will retrieve the username directly from the database(so won't be an argument of the constructor)

HugoHof: I will do this correctly in my next task, ask this time I have added a TODO comment

HugoHof: For simplicity, a userModel will be only created only after r the profile creation so we can considerate this username has correct

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
HugoHof: I don't think it will be useful because the userModel is always created with all this arguments at the end of the user profile creation

clementcharmillot (changes requested):
Thank for the fixes. Most of the issues are solved which is pretty good. I've still a few remarks though.

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
```diff
@@ -61,12 +80,25 @@ class FireDatabase : Database() {
         updateUserData(userAdd, username)
     }
 
+    override fun setDistanceGoal(userId: String, distanceGoal: Double): CompletableFuture<Boolean> {
+        TODO("Will be done in the next task: database rebase")
+    }
+
+    override fun setActivityTimeGoalGoal(userId: String, activityTimeGoal: Double): CompletableFuture<Boolean> {
```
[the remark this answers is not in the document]
HugoHof: Oh thanks, sorry

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
```diff
@@ -0,0 +1,312 @@
...
+
+    /**
+     * Use this function to modify the daily distance goal of the user
+     * @param distance new daily distance goal
+     */
+    fun setDistanceGoal(distance: Double): CompletableFuture<Boolean>{
+        if(distance <= 0.0){
+            throw java.lang.Error("The distance goal can't be equal to 0 !")
+        }
+        this.distanceGoal=distance
```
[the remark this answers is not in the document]
HugoHof: Thanks for the advice. I have updated the code.

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
```diff
@@ -0,0 +1,312 @@
...
+        this.nbOfPathsGoal=nbOfPaths
+        return database.setNbOfPathsGoalGoal(userId, nbOfPaths)
+    }
+
+    /**
+     * Get the age of the user
+     * @return the age of the user
+     */
+    fun getAge(): Int{
+        var age = LocalDate.now().year-dateOfBirth.year
```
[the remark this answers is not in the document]
HugoHof: Thanks for the advice, the code is more smaller with that

app/src/main/java/com/epfl/drawyourpath/database/MockDataBase.kt
```diff
@@ -31,13 +80,28 @@ class MockDataBase : Database() {
...
-        distanceGoal: Int,
-        timeGoal: Int,
+        distanceGoal: Double,
+        timeGoal: Double,
         nbOfPathsGoal: Int
     ) {
         val userGoals: UserGoals = UserGoals(username, distanceGoal, timeGoal, nbOfPathsGoal)
         usernameToUserGoalsDataBase.put(username, userGoals)
     }
+
```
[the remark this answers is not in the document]
HugoHof: It will refactor this class during theme next task, so  will probably delete this method to sendall the PersoInfo and the user goals in a same transaction.

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
```diff
@@ -0,0 +1,312 @@
...
+     * @param firstname must respect the name convention of the app (name or name-name)
+     * @param surname must respect the name convention of the app (name or name-name)
+     * @param dateOfBirth the user be aged between 10 and 100 years old
+     * @param distanceGoal init at the user profile creation and can be modify after(daily goal)
+     * @param activityTimeGoal init at the user profile creation and can be modify after(daily goal)
+     * @param nbOfPathsGoal init at the user profile creation and can be modify after(daily goal)
+     * @throws error if the inputs are incorrect
+     */
+    constructor(userAuth: User, firstname: String, surname: String, dateOfBirth: LocalDate, distanceGoal: Double, activityTimeGoal: Double, nbOfPathsGoal: Int, database: Database){
+        this.database = database
```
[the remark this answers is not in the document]
HugoHof: I have change the checking to evict code duplication. Thank you for your advice .

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
```diff
@@ -0,0 +1,362 @@
...
+     */
+    fun getUsername(): String{
+        return username
+    }
+
+    /**
+     * Use this function to modify the username(the username will be modify only if it is available on the database)
+     * @param username that we want to set
+     */
+    fun setUsername(username: String): CompletableFuture<Boolean>{
```
clementcharmillot, line 110: Should be 
```kotlin
        return database.updateUsername(username, userId).thenApply{
            if (it) {
                this.username=username
            }
            it
        }
```

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
```diff
@@ -0,0 +1,362 @@
...
+    }
+    return true
+}
+
+/**
+ * Helper function to check if the firstname format is correct and throw directly an error if it is incorrect
+ * @param firstanme to be check
+ * @return true if the format is correct and throw an error otherwise
+ */
+private fun checkFirstname(firstname: String): Boolean{
```
clementcharmillot, line 286: `checkFirstname` and `checkSurname` are exactly the same methods, they should be merged in a single one.

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
```diff
@@ -0,0 +1,362 @@
...
+    if(!checkNameFormat(firstname)){
+        throw java.lang.Error("Incorrect firstname")
+    }
+    return true
```
clementcharmillot, line 290: Why return true?

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
```diff
@@ -0,0 +1,362 @@
...
+/**
+ * Helper function to check if the date of birth of the user respect the age condition of the app
+ * @param date of the user birth
+ * @return return true if the user is aged between 10 and 100 years and throw an error otherwise
+ */
+private fun checkDateOfBirth(date: LocalDate): Boolean{
+    if(!(date < LocalDate.now().plusYears(-10) && date > LocalDate.now().plusYears(-100))){
+        throw java.lang.Error("Incorrect date of birth !")
+    }
+    return true
```
clementcharmillot, line 322: Why return true?

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
```diff
@@ -0,0 +1,362 @@
...
+/**
+ * Helper function to check if the distance goal is greater or equal than zero
+ * @param distanceGoal to be checked
+ * @return true if the goal is correct and throw an error otherwise
+ */
+private fun checkDistanceGoal(distanceGoal: Double): Boolean{
+    if(distanceGoal<=0.0){
+        throw java.lang.Error("The distance goal can't be equal or less than 0.")
+    }
+    return true
```
clementcharmillot, line 334: Same.

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
```diff
@@ -0,0 +1,362 @@
...
+/**
+ * Helper function to check if the number of paths goal is greater or equal than zero
+ * @param nbOfPathsGoal to be checked
+ * @return true if the goal is correct and throw an error otherwise
+ */
+private fun checkNbOfPathsGoal(nbOfPathsGoal: Int): Boolean{
+    if(nbOfPathsGoal<=0){
+        throw java.lang.Error("The number of paths goal can't be equal or less than 0.")
+    }
+    return true
```
clementcharmillot, line 358: Same.

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
```diff
@@ -0,0 +1,362 @@
...
+/**
+ * Helper function to check if the activity time goal is greater or equal than zero
+ * @param activityTimeGoal to be checked
+ * @return true if the goal is correct and throw an error otherwise
+ */
+private fun checkActivityTimeGoal(activityTimeGoal: Double): Boolean{
+    if(activityTimeGoal<=0.0){
+        throw java.lang.Error("The activity time goal can't be equal or less than 0.")
+    }
+    return true
```
clementcharmillot, line 346: Same.

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
```diff
@@ -0,0 +1,362 @@
...
+/**
+ * Helper function to check if the surname format is correct and throw directly an error if it is incorrect
+ * @param surname to be check
+ * @return true if the format is correct and throw an error otherwise
+ */
+private fun checkSurname(surname: String): Boolean{
+    if(!checkNameFormat(surname)){
+        throw java.lang.Error("Incorrect surname")
+    }
+    return true
```
clementcharmillot, line 302: Same.

clementcharmillot (changes requested):
A few bugs appeared due to the refactoring and after that we are good for me. Great job.

As the check functions throw on error, we don't need to return the boolean anymore.

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
```diff
@@ -0,0 +1,343 @@
...
+    }
+
+}
+
+/**
+ * Helper function to check if the firstname format is correct and throw directly an error if it is incorrect
+ * @param name to be check
+ * @throw an error if the format is not correct
+ */
+private fun checkFirstnameFormat(name: String){
```
clementcharmillot, line 273: `private fun checkFirstnameFormat(name: String)` and `private fun checkSurnameFormat(name: String)` are still the same function and code is duplicated. We can just have a single `checkNameFormat()` one I think.

clementcharmillot (approved):
Just one last thing and we can merge 👍. Don't forget to squash the commits when merging.

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
HugoHof: I think that if we just have a checkName function, we can't differentiate if the error come from the username or the first name

HugoHof: Thank for your review. I m waiting your answer for merging.

clementcharmillot: We could pass the checked property name as a string parameter and use it when throwing the exception.

PR #68, opened 2023-03-22
rebase the database: create a map username->userid and the the user a…

### LINKED ISSUE #49: Rebase the database such that the data of the user profile is linked to the userId(given by the login part) instead of the username.

HugoHof:
In this PR, I have rebase the database such that the user account on the database are directly linked to their userId (instead of their username). So now, the database contains two folds : 
- [usernameToUserId]: That contains links from username to userId. This is used to test if a username is available for init the username and also to update it. 
- [users] that contains the users account. Each user account is linked to a user id. At the initial state(after user profile creation), the user account have: ( a username(same as  usernameToUserId), an email(retrieve from auth part), a first name, a surname, a dateOfBirth, a distance goal, a activity time goal and a number of path goal). 

I have also added in the database files some helper function to facilitate the used of the database :
- a method to know of a user with userId is tore to the database
- a method to obtain the username of a username for his userId
- a method to obtain the userId of a user from his username
- a method to know if a username is available
- a method to update the username of the logged user
- a method to set at the profile creation the username (and in a same time create the user account)
- a method to init the user account with first name, surname, dateOfBirth and the distance, time and nb of paths goals. This method used a UserModel as parameter and his used at the end of the profile creation(to make one transaction of data instead of two before).
- a method to get the userAccount (a userModel of this account) with a userId
- a method to get the userModel of the user logged in the app
- one method to update the distance goal of the user logged in the app
- one method to update the activity time goal of the user logged in the app
- one method to update the number of paths goal of the user logged in the app

Like before there is one Firebase database for the current usage of the app and one MockDatabase to simulate the database for the tests.

Tank you in advance for your review 
close #49


clementcharmillot:
Is it ok for you if I review this PR as well as it concerns the database?

HugoHof:
> Is it ok for you if I review this PR as well as it concerns the database?
> EDIT: PR got moved to draft so maybe after it's ready.

Yes you can if you want

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
```diff
@@ -38,76 +70,175 @@ class FireDatabase : Database() {
...
     }
 
-    override fun setDistanceGoal(userId: String, distanceGoal: Double): CompletableFuture<Boolean> {
-        TODO("Will be done in the next task: database rebase")
+    override fun setDistanceGoal(distanceGoal: Double): CompletableFuture<Boolean> {
+        val future = CompletableFuture<Boolean>()
+        val dataUpdated = HashMap<String, Any>()
+        dataUpdated.put("distanceGoal", distanceGoal)
+
+        if(distanceGoal <= 0.0){
```
clementcharmillot, line 189: The error should be handle through the future.

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
```diff
@@ -38,76 +70,175 @@ class FireDatabase : Database() {
...
+                    future.completeExceptionally(java.lang.Error("There is no user logged on the app."))
+                }else{
+                    val firstname = userData.child("firstname").value
+                    val surname = userData.child("surname").value
+                    val dateOfBirth = userData.child("dateOfBirth").value
+                    val distanceGoal = userData.child("distanceGoal").value
+                    val activityTimeGoal = userData.child("activityTimeGoal").value
+                    val nbOfPathsGoal = userData.child("nbOfPathsGoal").value
+                    if(firstname==null||surname==null||dateOfBirth==null||distanceGoal==null||activityTimeGoal==null||nbOfPathsGoal==null){
+                        future.completeExceptionally(java.lang.Error("The user account present on the database is incomplete."))
```
clementcharmillot, line 165: I think we should allow user account to miss some data for example the goals or the birth date.

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
```diff
@@ -38,76 +70,175 @@ class FireDatabase : Database() {
...
+        val future = CompletableFuture<UserModel>()
+
+        database.child(usersProfileFileName).child(userId).get().addOnSuccessListener { userData ->
+            if(userData == null) {
+                future.completeExceptionally(java.lang.Error("There is no user account corresponding to this userId."))
+            }else{
+                if(userAuth==null){
+                    future.completeExceptionally(java.lang.Error("There is no user logged on the app."))
+                }else{
+                    val firstname = userData.child("firstname").value
```
clementcharmillot, line 158: The strings to access properties are duplicated across the file. Making a rename of the fields pretty hard in the future. They should be constant variable.

app/src/main/java/com/epfl/drawyourpath/userProfileCreation/UserGoalsInitFragment.kt
```diff
@@ -28,7 +35,10 @@ class UserGoalsInitFragment : Fragment(R.layout.fragment_user_goals_init) {
             isTest = false
         } else {
             isTest = argsFromLastFrag.getBoolean("isRunningTestForDataBase")
-            username = argsFromLastFrag.getString("userName").toString()
+            username = argsFromLastFrag.getString("username").toString()
```
clementcharmillot, line 38: Use global variable instead of string literals as said above.

app/src/main/java/com/epfl/drawyourpath/userProfileCreation/UserGoalsInitFragment.kt
```diff
@@ -78,18 +89,27 @@ class UserGoalsInitFragment : Fragment(R.layout.fragment_user_goals_init) {
...
-                    fragManagement.replace(R.id.userGoalInitFragment, endProfileCreationFrag)
-                        .commit()
+                var userLog = FirebaseAuth.getUser()
+                if(isTest){
+                    userLog = MockAuth.MOCK_USER
+                }
+                if(userLog == null) {
+                    throw java.lang.Error("Any user is logged to the app.")
+                }
+                val user = UserModel(userLog, firstname, surname, LocalDate.ofEpochDay(dateOfBirth), distanceGoal.toDouble(),
```
clementcharmillot, line 99: There is a problem with how `UserModel` works here. The constructor isn't trivial as it fetches the database for the username. I think that we should modify the constructor to take the username so it doesn't fetch the database whenever it is created.

app/src/main/java/com/epfl/drawyourpath/userProfileCreation/UserGoalsInitFragment.kt
```diff
@@ -78,18 +89,27 @@ class UserGoalsInitFragment : Fragment(R.layout.fragment_user_goals_init) {
...
+                    timeGoal.toDouble(), nunberOfPathGoal, database)
+                database.initUserProfile(user).thenAccept{
+                    val previousActivity = activity
```
clementcharmillot, line 102: Why store the current activity in an intermediate variable?

app/src/main/java/com/epfl/drawyourpath/userProfileCreation/UserGoalsInitFragment.kt
```diff
@@ -78,18 +89,27 @@ class UserGoalsInitFragment : Fragment(R.layout.fragment_user_goals_init) {
...
-                    val endProfileCreationFrag = EndProfileCreationFragment()
-                    endProfileCreationFrag.arguments = dataToEndProfileCreationFrag
-                    fragManagement.replace(R.id.userGoalInitFragment, endProfileCreationFrag)
-                        .commit()
+                var userLog = FirebaseAuth.getUser()
+                if(isTest){
+                    userLog = MockAuth.MOCK_USER
+                }
+                if(userLog == null) {
+                    throw java.lang.Error("Any user is logged to the app.")
```
clementcharmillot, line 97: Throwing in a fragment will crash the app. Here, if the user isn't logged, we don't want to terminate the app but to redirect to the login activity.

app/src/main/java/com/epfl/drawyourpath/userProfileCreation/UserNameTestAndSetFragment.kt
```diff
@@ -52,16 +52,19 @@ class UserNameTestAndSetFragment : Fragment(R.layout.fragment_user_name_test_and
...
-                    //data to transmit to the PersonalInfoFragment(username + isTest)
-                    dataToPersoInfoFrag.putBoolean("isRunningTestForDataBase", isTest)
-                    dataToPersoInfoFrag.putString("userName", inputUserName.text.toString())
-                    val persoInfoFrag = PersonalInfoFragment()
-                    persoInfoFrag.arguments = dataToPersoInfoFrag
-                    fragManagement.replace(R.id.userName_frame, persoInfoFrag).commit()
+                if (previousActivity != null) {
+                    database.setUsername(usernameStr).thenAccept{isSet ->
+                        if(isSet){
+                            val fragManagement = previousActivity.supportFragmentManager.beginTransaction()
```
clementcharmillot, line 58: Could be extracted to a method called `showPersonalInfoFragment` to make the code more readable.
It isnt' clear at first glance what it does.

app/src/main/java/com/epfl/drawyourpath/userProfileCreation/UserGoalsInitFragment.kt
```diff
@@ -78,18 +89,27 @@ class UserGoalsInitFragment : Fragment(R.layout.fragment_user_goals_init) {
...
+                }
+                val user = UserModel(userLog, firstname, surname, LocalDate.ofEpochDay(dateOfBirth), distanceGoal.toDouble(),
+                    timeGoal.toDouble(), nunberOfPathGoal, database)
+                database.initUserProfile(user).thenAccept{
+                    val previousActivity = activity
+                    if (previousActivity != null) {
+                        val fragManagement = previousActivity.supportFragmentManager.beginTransaction()
```
clementcharmillot, line 104: The code to switch fragment is duplicated along all profile creation fragments. There should be a super class for all profile creation fragments that handles fragment transition.

app/src/main/java/com/epfl/drawyourpath/userProfileCreation/PersonalInfoFragment.kt
```diff
@@ -111,14 +111,16 @@ class PersonalInfoFragment : Fragment(R.layout.fragment_personal_info) {
...
             //if the data are correct, set them to the database and show the next fragment when click on the validate button
             if (test1 && test2 && test3) {
-                database.setPersonalInfo(userName, firstname, surname, dateOfBirth)
                 val previousActivity = activity
                 if (previousActivity != null) {
                     val fragManagement = previousActivity.supportFragmentManager.beginTransaction()
                     val dataToUserGoalsInitFrag: Bundle = Bundle()
-                    //data to transmit to the UserGoalsInitFragment(username + isTest)
+                    //data to transmit to the UserGoalsInitFragment(username + firstname + surname + dateOfBirth + isTest)
                     dataToUserGoalsInitFrag.putBoolean("isRunningTestForDataBase", isTest)
```
clementcharmillot, line 119: Same remark concerning the keys.

app/src/main/java/com/epfl/drawyourpath/database/MockDataBase.kt
```diff
@@ -46,73 +53,79 @@ class MockDataBase : Database() {
...
+        userIdToUserAccount.put(userIdTest, updatedUser)
+        return CompletableFuture.completedFuture(true)
     }
 
-    override fun setNbOfPathsGoal(userId: String, nbOfPathsGoal: Int): CompletableFuture<Boolean> {
-        userIdToNbOfPathsGoal.put(userId, nbOfPathsGoal)
-        return CompletableFuture.completedFuture(userIdToNbOfPathsGoal.get(userId) == nbOfPathsGoal)
+    override fun setNbOfPathsGoal(nbOfPathsGoal: Int): CompletableFuture<Boolean> {
+        if(nbOfPathsGoal <= 0.0){
+            throw java.lang.Error("The number of paths goal can't be less or equal than 0.")
```
clementcharmillot, line 124: Error should be thrown through the future.

app/src/main/java/com/epfl/drawyourpath/database/MockDataBase.kt
```diff
@@ -46,73 +53,79 @@ class MockDataBase : Database() {
...
+        userIdToUserAccount.put(userIdTest, updatedUser)
+        return CompletableFuture.completedFuture(true)
     }
 
-    override fun setActivityTimeGoal(userId: String, activityTimeGoal: Double): CompletableFuture<Boolean> {
-        userIdToActivityTimeGoal.put(userId, activityTimeGoal)
-        return CompletableFuture.completedFuture(userIdToActivityTimeGoal.get(userId) == activityTimeGoal)
+    override fun setActivityTimeGoal(activityTimeGoal: Double): CompletableFuture<Boolean> {
+        if(activityTimeGoal <= 0.0){
+            throw java.lang.Error("The activity time goal can't be less or equal than 0.")
```
clementcharmillot, line 114: Error should be thrown through the future.

app/src/main/java/com/epfl/drawyourpath/database/MockDataBase.kt
```diff
@@ -46,73 +53,79 @@ class MockDataBase : Database() {
...
+    override fun getLoggedUserAccount(): CompletableFuture<UserModel> {
+        return getUserAccount(userIdTest)
     }
 
-    override fun setDistanceGoal(userId: String, distanceGoal: Double): CompletableFuture<Boolean> {
-        userIdToDistanceGoal.put(userId, distanceGoal)
-        return CompletableFuture.completedFuture(userIdToDistanceGoal.get(userId) == distanceGoal)
+    override fun setDistanceGoal(distanceGoal: Double): CompletableFuture<Boolean> {
+        if(distanceGoal <= 0.0){
+            throw java.lang.Error("The distance goal can't be less or equal than 0.")
```
clementcharmillot, line 104: Error should be thrown through the future.

app/src/main/java/com/epfl/drawyourpath/database/MockDataBase.kt
```diff
@@ -46,73 +53,79 @@ class MockDataBase : Database() {
...
     }
 
-    override fun setActivityTimeGoal(userId: String, activityTimeGoal: Double): CompletableFuture<Boolean> {
-        userIdToActivityTimeGoal.put(userId, activityTimeGoal)
-        return CompletableFuture.completedFuture(userIdToActivityTimeGoal.get(userId) == activityTimeGoal)
+    override fun setActivityTimeGoal(activityTimeGoal: Double): CompletableFuture<Boolean> {
+        if(activityTimeGoal <= 0.0){
+            throw java.lang.Error("The activity time goal can't be less or equal than 0.")
+        }
+        val updatedUser = UserModel(userAuthTest, firstnameTest, surnameTest, dateOfBirthTest, distanceGoalTest,
```
clementcharmillot, line 116: Same problem here concerning the `UserModel`, the mock database ends up doing fetches on the firebase backend.

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
```diff
@@ -38,76 +70,175 @@ class FireDatabase : Database() {
...
-        database.child("users").child(username).updateChildren(data)
-            .addOnFailureListener {
-                throw Exception(buildString {
-                    append("Impossible to add the data on the user: ")
-                    append(username)
-                    append(" on the database")
-                })
-            }
+    private fun getUserId(): String{
+        if(userAuth == null){
```
clementcharmillot, line 235: Throwing an exception is maybe a bit overkill here. Couldn't we return `null`?

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
```diff
@@ -38,76 +70,175 @@ class FireDatabase : Database() {
...
+    override fun setDistanceGoal(distanceGoal: Double): CompletableFuture<Boolean> {
+        val future = CompletableFuture<Boolean>()
+        val dataUpdated = HashMap<String, Any>()
+        dataUpdated.put("distanceGoal", distanceGoal)
+
+        if(distanceGoal <= 0.0){
+            throw java.lang.Error("The distance goal can't be less or equal than 0.")
+        }
+
+        database.child(usersProfileFileName).child(getUserId()).updateChildren(dataUpdated)
```
clementcharmillot, line 193: `database.child(usersProfileFileName).child(getUserId()).updateChildren(dataUpdated)` is copied 3 times in this file. It should be refactored to a method `updateUser(uid: String, data: Map<String, Any>)`.

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
```diff
@@ -38,76 +70,175 @@ class FireDatabase : Database() {
...
+        userData.put("surname", userModel.getSurname())
+        userData.put("dateOfBirth", userModel.getDateOfBirth().toEpochDay())
+        userData.put("distanceGoal", userModel.getDistanceGoal())
+        userData.put("activityTimeGoal", userModel.getActivityTime())
+        userData.put("nbOfPathsGoal", userModel.getNumberOfPathsGoal())
+
+        val future = CompletableFuture<Boolean>()
+        val userId = getUserId()
+
+        database.child(usersProfileFileName).child(userId).updateChildren(userData)
```
clementcharmillot, line 142: `database.child(usersProfileFileName).child(userId)` is copied 4 times in this file, it should be refactored to its own function.

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
```diff
@@ -38,76 +70,175 @@ class FireDatabase : Database() {
...
+                    val surname = userData.child("surname").value
+                    val dateOfBirth = userData.child("dateOfBirth").value
+                    val distanceGoal = userData.child("distanceGoal").value
+                    val activityTimeGoal = userData.child("activityTimeGoal").value
+                    val nbOfPathsGoal = userData.child("nbOfPathsGoal").value
+                    if(firstname==null||surname==null||dateOfBirth==null||distanceGoal==null||activityTimeGoal==null||nbOfPathsGoal==null){
+                        future.completeExceptionally(java.lang.Error("The user account present on the database is incomplete."))
+                    }else{
+                        future.complete(
+                            UserModel(userAuth, firstname as String, surname as String, LocalDate.ofEpochDay(dateOfBirth as Long),
```
clementcharmillot, line 168: We end up fetching the username two times due to the way the `UserModel` constructor works.

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
```diff
@@ -38,76 +70,175 @@ class FireDatabase : Database() {
...
-
-        //for the date
-        val unixDate: Long = dateOfBirth.toEpochDay()
-        val userDateMap = HashMap<String, Long>()
-        userDateMap.put("dateOfBirth", unixDate)
-        updateUserData(userDateMap, username)
+    override fun getUserAccount(userId: String): CompletableFuture<UserModel> {
+        val future = CompletableFuture<UserModel>()
+
+        database.child(usersProfileFileName).child(userId).get().addOnSuccessListener { userData ->
```
clementcharmillot, line 151: I think there is a problem with how data is stored. Here we fetch all the data present at `users/{uid}/`. We should break down the data in the database in smalled chunks. For example, we could create a `users/{uid}/public` node to only fetch profile information to init the view model. It would prevent fetching the entire data of the user.

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
```diff
@@ -38,76 +70,175 @@ class FireDatabase : Database() {
...
+        }
+        return future
+    }
+
+    override fun setUsername(username: String): CompletableFuture<Boolean> {
+        val future = CompletableFuture<Boolean>()
+
+        val userId = getUserId()
+        //check if the username is available
+        isUsernameAvailable(username).thenAccept{isAvailable ->
```
clementcharmillot, line 103: Username availability shouldn't be handled on the client but through Firebase Security Rules.

This way, we can only try to set it without checking its availability first and it will fail due to the security rules.

We could let it like that for now but add a `TODO` for the future.

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
```diff
@@ -23,13 +35,33 @@ class FireDatabase : Database() {
...
+            future.completeExceptionally(it)
+        }
+        return future
     }
 
     override fun isUsernameAvailable(userName: String): CompletableFuture<Boolean> {
         val future = CompletableFuture<Boolean>()
 
-        database.child("users").child(userName).get().addOnSuccessListener {
+        database.child(usernameToUserIdFileName).child(userName).get().addOnSuccessListener {
```
clementcharmillot, line 64: `database.child(usernameToUserIdFileName)` is copied 4 times in this file. It should be refactored.

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
```diff
@@ -38,76 +70,175 @@ class FireDatabase : Database() {
...
 
-    override fun setNbOfPathsGoal(userId: String, nbOfPathsGoal: Int): CompletableFuture<Boolean> {
-        TODO("Will be done in the next task: database rebase")
+    override fun setNbOfPathsGoal(nbOfPathsGoal: Int): CompletableFuture<Boolean> {
+        val future = CompletableFuture<Boolean>()
+        val dataUpdated = HashMap<String, Any>()
+        dataUpdated.put("nbOfPathsGoal", nbOfPathsGoal)
+
+        if(nbOfPathsGoal <= 0.0){
+            throw java.lang.Error("The number of paths goal can't be less or equal than 0.")
```
clementcharmillot, line 220: Error should be thrown through the future.

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
```diff
@@ -38,76 +70,175 @@ class FireDatabase : Database() {
...
 
-    override fun setActivityTimeGoal(userId: String, activityTimeGoal: Double): CompletableFuture<Boolean> {
-        TODO("Will be done in the next task: database rebase")
+    override fun setActivityTimeGoal(activityTimeGoal: Double): CompletableFuture<Boolean> {
+        val future = CompletableFuture<Boolean>()
+        val dataUpdated = HashMap<String, Any>()
+        dataUpdated.put("activityTimeGoal", activityTimeGoal)
+
+        if(activityTimeGoal <= 0.0){
+            throw java.lang.Error("The activity time goal can't be less or equal than 0.")
```
clementcharmillot, line 205: Error should be thrown through the future.

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
```diff
@@ -8,12 +11,21 @@ import java.util.concurrent.CompletableFuture
...
+
+/**
+ * The Firebase contains files:
+ * -usernameToUserId: that link the username to a unique userId
+ * -users: that contains users based on the UserModel defined by their userId
+ */
 class FireDatabase : Database() {
     val database: DatabaseReference = Firebase.database.reference
+    private val userAuth: User? = FirebaseAuth.getUser()
+    private val usernameToUserIdFileName: String = "usernameToUserId"
```
clementcharmillot, line 23: Should be extracted as constants and follow casing convention of constants (all caps).

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
```diff
@@ -38,76 +70,175 @@ class FireDatabase : Database() {
...
+            if(isAvailable){
+                //add the link between the username and the userId
+                val usernameToUserId = HashMap<String, String>()
+                usernameToUserId.put(username, userId)
+                database.child(usernameToUserIdFileName).updateChildren(usernameToUserId as Map<String, Any>)
+                    .addOnSuccessListener {
+                        //add the users account to the database and the username to the user account
+                        val userAccount = HashMap<String, String>()
+                        userAccount.put("username", username)
+                        database.child("users").child(userId).updateChildren(userAccount as Map<String, Any>)
```
clementcharmillot, line 113: `database.child("users")` doesn't use the `usersProfileFileName` constant.

clementcharmillot (changes requested):
Very important PR that will make the link between the local user data and the database. I think it prepares well for the upcoming features.

Just a few remarks concerning the code and the implementation.

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
HugoHof: I think we shouldn't do that because in the create profile account activity the user is obligate to enter all this information. So if we don't have all this information then there is an error in the profile creation(and we shouldn't have that).

app/src/main/java/com/epfl/drawyourpath/userProfileCreation/UserGoalsInitFragment.kt
HugoHof: Since, the user creation activity is launch just after the login, there must be a user logged in the app. If it's not case, there is an error in the logging protocol so we must launch an error.

app/src/main/java/com/epfl/drawyourpath/userProfileCreation/UserGoalsInitFragment.kt
HugoHof: The problem is that the arguments transmitted between the fragment are not the same along the profile creation, so it will be difficult to create a commun function.

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
clementcharmillot: We might assume that we don't ask the age of the user for some users based on confidentiality. But it might be changed later.

app/src/main/java/com/epfl/drawyourpath/userProfileCreation/UserGoalsInitFragment.kt
clementcharmillot: The SDK can logs the user out for a few reasons and `getUser()` will return `null` so it doesn't necesseraly means there is an error in the login activity. You are right that an error should be recorded but crashing the app isn't a user-friendly approach.

app/src/main/java/com/epfl/drawyourpath/userProfileCreation/UserGoalsInitFragment.kt
clementcharmillot: You could allow the bundle as an argument to pass any data:
```kotlin
fun switchFragment<T>(fragId: Int, bundle: Bundle = Bundle()) {
    val fragManagement activity.supportFragmentManager.beginTransaction()
    fragManagement.replace(fragId, T()).commit()
}
```

app/src/main/java/com/epfl/drawyourpath/database/MockDataBase.kt
HugoHof: resolve: we dont't use the database inside the init user model

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
HugoHof: We couldn't because we need a non null userid in the functions that that call this function.

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
HugoHof: I think we could do that in future task when we will choose the privacy rules

clementcharmillot: It isn't really linked to privacy, as now you would fetch all the races, etc. of the user just to get basic data.

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
clementcharmillot: I see what you mean, but there are two main problem: the user id is fetched upon the creation of the UserModel which means that we could reference a stale uid. All methods using this function are asynchronous and return a future. If this function starts throwing, the exception won't be forwarded through the future and will just crash the app.

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
HugoHof: At this time all the information store in the userModel are useful. I don't really understand what you want to do.

app/src/main/java/com/epfl/drawyourpath/userProfileCreation/UserGoalsInitFragment.kt
HugoHof: Ok I will change for launching the login activity it's done

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
```diff
@@ -38,76 +71,199 @@ class FireDatabase : Database() {
...
+        //TODO:Add security rules to database
+        val userId = getUserId()
+        if(userId == null){
+            future.completeExceptionally(java.lang.Error("The userId can't be null !"))
+        }else{
+            //check if the username is available
+            isUsernameAvailable(username).thenAccept { isAvailable ->
+                if (isAvailable) {
+                    //add the link between the username and the userId
+                    val usernameToUserId = HashMap<String, String>()
```
clementcharmillot, line 120: Doing two transactions here could result in partial state being saved to the database in case of connection failure. Realtime Database allows to do that in a single operation by using a complete path of the data.

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
```diff
@@ -38,76 +71,199 @@ class FireDatabase : Database() {
...
+        val future = CompletableFuture<UserModel>()
+
+        database.child(usersProfileFileName).child(userId).get().addOnSuccessListener { userData ->
+            if(userData == null) {
+                future.completeExceptionally(java.lang.Error("There is no user account corresponding to this userId."))
+            }else{
+                if(USER_AUTH==null){
+                    future.completeExceptionally(java.lang.Error("There is no user logged on the app."))
+                }else{
+                    val username = userData.child(usernameFile).value
```
clementcharmillot, line 168: It makes the function quite complex. Might be worth it to refactor in a `userFileToUserModel` function maybe?

clementcharmillot (changes requested):
Thanks for the changes, it solves pretty much all my remarks. I just have a few more remarks mostly about making the code a bit easier to read.

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
```diff
@@ -38,76 +71,199 @@ class FireDatabase : Database() {
         return future
     }
 
-    override fun updateUsername(username: String, userId: String): CompletableFuture<Boolean> {
-        TODO("This will be implmented during next task when I will clean the database organisation")
+    override fun updateUsername(username: String): CompletableFuture<Boolean> {
```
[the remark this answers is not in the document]
HugoHof: But I will add an helper function to remove the past link username->userId

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
```diff
@@ -38,76 +71,199 @@ class FireDatabase : Database() {
...
+                    future.completeExceptionally(java.lang.Error("Impossible to find the past username !"))
+                } else {
+                    //update the link username to userId and the username on the userAccount
+                    setUsername(username).thenAccept { isSetUsername ->
+                        if (isSetUsername) {
+                            //remove the past username from the link username/userId
+                            accessUsernameToUserIdFile().child(pastUsername).removeValue()
+                                .addOnSuccessListener { future.complete(true) }
+                                .addOnFailureListener {
+                                    future.completeExceptionally(
```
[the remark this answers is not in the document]
HugoHof: It's t important to know if it fail, because if it's not delete correctly the username will be present as taken but it is free.

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
HugoHof: I don't think it will be useful since this code is only present in this function. So create an helper wouldn't have a great interest.

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
HugoHof: I understand but we want to make sur that the previous transaction its made bedaiore begin a new one. And in any case an error will be throw in this case

HugoHof:
> Thanks for the changes, it solves pretty much all my remarks. I just have a few more remarks mostly about making the code a bit easier to read.

Thank you for your review. I have implemented the most of your comments

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
clementcharmillot: It would make the function a lot more readable.

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
clementcharmillot: But then shouldn't we rollback the update? Because the name will be updated in the profile but not in the mapping.

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
clementcharmillot: It would give the function below
```kotlin
override fun getUserAccount(userId: String): CompletableFuture<UserModel> {
    val future = CompletableFuture<UserModel>()

    accessUserAccountFile(userId).get().addOnSuccessListener { userData ->
        val userData = fileToUserModel(userData)
        if (userData != null) {
            future.complete(userData)
        }
        else {
            future.completeExceptionally(it)
        }
    }.addOnFailureListener{
        future.completeExceptionally(it)
    }
    return future
}
```

HugoHof: It's done thanks

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
clementcharmillot: After further discussion, we decided to go that way for now and implement the other method at a later time.

clementcharmillot (approved):
Thanks for addressing my remarks. All good for me we can merge now.

PR #70, opened 2023-03-26
56 tournaments UI

### LINKED ISSUE #56: create UI for displaying the tournaments in which the user is in or can take

HoLuDyEr:
create the UI for the tournaments in the community fragment

when launching the community fragment there are some posts that are displayed.

image

The user can then scroll and upvote/downvote each posts.

image

When pressing on the top left, it opens a drawer containing 3 submenus, the first one is the weekly tournament, the second one the tournament in which the user participate, lastly some discover tournaments that are proposed to the user.

image

When selecting a tournament, it displays some information about it and displays only posts of this tournament.

image

the backend of the tournament are currently filled with placeholder and needs to be changed so that it can use the database.

also there is some modification of the build.gradle file as there was a bug that appear only on API 28-30 (CI emulator has API 30) (https://github.com/android/android-test/issues/1642) so I changed the version of espresso to 3.5.0-alpha04. Changing this version will result in the tests failing.

closes #56


damiengenoud:
Great UI, I find it clear and easy to navigate !

I do have one small remark before reviewing the code itself: on some device (like the Pixel 2) while navigating the submenus, the "description" and starting time take a bit too much space, to the point where the username and votes cannot be seen at the same time, see for example (Pixel 2):

Capture d’écran 2023-03-26 225019

I think it would be better, if possible, to have the bar containing the tournament description being thiner, as it would fit on more device.

Another possibility would be to remove the "drawers" button on the bottom right of the bar, as it doesn't seem useful for now.

app/src/main/java/com/epfl/drawyourpath/community/Tournament.kt
```diff
@@ -1,5 +1,6 @@
-package com.epfl.drawyourpath.challenge
+package com.epfl.drawyourpath.community
 
+import com.epfl.drawyourpath.community.TournamentPost
```
damiengenoud, line 3: Detail: this import does not seem to be useful.

app/src/main/java/com/epfl/drawyourpath/mainpage/fragments/CommunityFragment.kt
```diff
@@ -1,8 +1,258 @@
...
-    // TODO: add community UI here (like rating/competition) and the ViewModel for this screen/fragment.
+
+    private lateinit var drawer: DrawerLayout
+    private lateinit var menuButton: ImageButton
+    private lateinit var detailsLayout: LinearLayout
+    private lateinit var tournamentsView: RecyclerView
+
+    private lateinit var tournament: TournamentModel
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
```
damiengenoud, line 36: This function could be improved in readability by factoring it even more in smaller functions in my opinion.

app/src/main/java/com/epfl/drawyourpath/mainpage/fragments/CommunityFragment.kt
```diff
@@ -1,8 +1,258 @@
...
+                }
+        }
+        val discover = menu.addSubMenu("Discover")
+        for (i in tournament.getDiscoverTournament("placeholder")) {
+            discover.add(i.name)
+                .setContentDescription("${i.name} details")
+                .setOnMenuItemClickListener {
+                    menuItemListener(view, i)
+                }
+        }
```
damiengenoud, line 99: These three block of codes are very similar, it should be possible to factorize them with a private function.

app/src/androidTest/java/com/epfl/drawyourpath/mainpage/fragments/CommunityFragmentTest.kt
damiengenoud, line 1: Good testing ! 
Just a detail: I think the naming of the tests could be improved by indicating what the expected output is, instead of ending with "...IsCorrect()", to have a better idea of the problem without needing to read the tests. The comments that explain this could then be directly in the test name.

Example: upvoteOnceIsCorrect() with comment "test that pressing the upvote button adds a vote"    ->
upvoteOnceAddsAVote(), or something similar.

damiengenoud (changes requested):
Very good PR with a lot of code that will be very helpful when implementing the social part of the app, good job ! There are just a few very small improvements that could be made in the code, and a displaying issue on small screens, but otherwise it seems all good !

app/src/main/java/com/epfl/drawyourpath/community/TournamentPost.kt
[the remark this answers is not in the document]
HoLuDyEr: yes, it is a better idea. Thank you for pointing this out.

app/src/main/java/com/epfl/drawyourpath/community/Tournament.kt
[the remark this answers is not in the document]
HoLuDyEr: Nice catch

app/src/main/java/com/epfl/drawyourpath/mainpage/fragments/CommunityFragment.kt
HoLuDyEr: I refactored it, now it is easier to understand

app/src/main/java/com/epfl/drawyourpath/mainpage/fragments/CommunityFragment.kt
HoLuDyEr: It is now inside a private function

app/src/androidTest/java/com/epfl/drawyourpath/mainpage/fragments/CommunityFragmentTest.kt
HoLuDyEr: Thank you for pointing this out. I renamed the test to more easier to understant names.

HoLuDyEr:
Thank you for the review, I took the comments into account.
Also I changed the UI as you mentioned that the description and date took too much space, they are now in a nested scroll view so only the back button and the tournament name stays at the top of the screen when scrolling the posts.

> Great UI, I find it clear and easy to navigate !
> 
> I do have one small remark before reviewing the code itself: on some device (like the Pixel 2) while navigating the submenus, the "description" and starting time take a bit too much space, to the point where the username and votes cannot be seen at the same time, see for example (Pixel 2):
> 
> Capture d’écran 2023-03-26 225019
> 
> I think it would be better, if possible, to have the bar containing the tournament description being thiner, as it would fit on more device.
> 
> Another possibility would be to remove the "drawers" button on the bottom right of the bar, as it doesn't seem useful for now.

damiengenoud (approved):
Well done for the nested scroll view, it is much easier to navigate with smaller screens! Thank you for the small fixes, it seems all good to me now!

PR #72, opened 2023-03-28
61 basic notifications

### LINKED ISSUE #61: Implement a basic notifications system

damiengenoud:
Adds a daily notification that reminds users to complete their challenges of the day. 

It does not yet take into account the fact that the user could have already completed the challenges, so the notification "asks" if the user has done it.

For now, the notification is displayed every day at 16:00, but it would be possible to change this in the future by adding a setting for the time of the reminder.

Clicking on the notification launches the login activity, so if the user is already logged in, the login activity directly redirects to the main activity, otherwise the user needs to authenticate.

The icon on the top bar when the notifications is displayed:
Capture d’écran 2023-03-28 041259

The notification itself:
Capture d’écran 2023-03-28 041607

closes #61


app/src/main/java/com/epfl/drawyourpath/notifications/NotificationsHelper.kt
```diff
@@ -0,0 +1,50 @@
...
+        }
+        remindersManager.startReminder(context)
+    }
+
+    private fun setupNotificationsChannels() {
+        //Setup challenges reminder channel:
+        setupChannel(
+            context.resources.getString(R.string.channel_challenges_reminder_id),
+            context.resources.getString(R.string.channel_challenges_reminder_name),
+            context.resources.getString(R.string.channel_challenges_reminder_desc),
```
HoLuDyEr, line 36: nice use of the strings values

app/src/main/java/com/epfl/drawyourpath/notifications/AlarmReceiver.kt
```diff
@@ -0,0 +1,74 @@
...
+        val notificationManager = ContextCompat.getSystemService(
+            context,
+            NotificationManager::class.java
+        ) as NotificationManager
+
+        val disableAutoReconnection = intent.getBooleanExtra(USE_MOCK_CHALLENGE_REMINDER, false)
+
+        notificationManager.sendChallengeReminderNotification(context, !disableAutoReconnection)
+        // Send other time-based notifications here
+    }
```
HoLuDyEr, line 34: Question: Does this function know which notification to send when the alarm is received or will it launch ervery notifications every time an alarm is received? 
(in the case where we add new notifications and add new reminders)

app/src/main/java/com/epfl/drawyourpath/notifications/AlarmReceiver.kt
```diff
@@ -0,0 +1,74 @@
...
+}
+
+/**
+ * Sends a challenge reminder notification
+ * @param applicationContext: the application context
+ */
+fun NotificationManager.sendChallengeReminderNotification(
+    applicationContext: Context, enableAutoReconnection: Boolean
+) {
```
HoLuDyEr, line 43: This function could be split so that parts of it can be reused by other notifications.

HoLuDyEr (changes requested):
It is a good PR it only requires small changes. The code is clean and well documented, well done.
I also have a few questions that I asked in the review.

app/src/main/res/values/strings.xml
```diff
@@ -56,4 +56,9 @@
     <string name="cancel">Cancel</string>
     <string name="apply">Apply</string>
     <string name="preferences_menu_item">Preferences</string>
+    <string name="channel_challenges_reminder_id">CHALLENGES</string>
+    <string name="channel_challenges_reminder_desc">Notifications to remind you to complete your weekly challenges.</string>
+    <string name="channel_challenges_reminder_name">Challenges Reminder</string>
+    <string name="challenge_reminder_notification_text">Have you completed all your challenges for today? If not, consider drawing more paths soon!</string>
+    <string name="challenge_reminder_notification_title">Challenges Reminder</string>
```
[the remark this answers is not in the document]
damiengenoud: Good catch, I merged both in one string "challenges_reminder"

app/src/main/res/xml/preferences.xml
```diff
@@ -34,14 +34,4 @@
...
-    <PreferenceCategory
-        app:key="notifications_category"
-        app:title="Notifications">
-
-        <SwitchPreferenceCompat
-            app:key="switch_notifications"
-            app:title="Enable notifications" />
-
-    </PreferenceCategory>
-
```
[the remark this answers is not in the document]
damiengenoud: I did it because this setting already exists in the settings app of the phone, where you can choose how the notifications for this app are received (sound, vibration, ...) and which one to enable or disable (due to the notifications channels). It might be possible to implement it directly in the app but it should not be a priority in my opinion, as it would merely be a "shortcut" to avoid going in the settings app.

app/src/main/java/com/epfl/drawyourpath/notifications/AlarmReceiver.kt
damiengenoud: I can split this function in smaller ones to be more readable, this is for sure a good idea, but I do not see how to make reusable functions out of it, as most (if not all) the code is specific for a challenge reminder notification and not for notifications in general. I still modularized it a bit, for readability mainly, thanks for pointing out this function !

damiengenoud:
Thanks for the review, I made the changes you requested as best as I could. I also simplified a test (clickingOnChallengeReminderNotificationLaunchesLoginActivity())for it to be more stable, and modularized startReminder() in DailyReminderManager.kt, as it was too long according to CodeClimnate (and I agreed).

app/src/main/res/xml/preferences.xml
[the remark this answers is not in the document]
HoLuDyEr: Thank you for the answer

HoLuDyEr:
Thank you for the changes and for answering my questions.
It looks good to me, you can merge now.

HoLuDyEr (approved):

clementcharmillot:
@damiengenoud I think you forgot to link this PR to #61

PR #78, opened 2023-03-30
Fix a bug in FireBase

HugoHof:
This PR fix a bug in the getLoggedUserAccount() function in the database. The future was never complete.

Thank to @SaftaKrist to find this problem.


CristianSafta (approved):
Approved

PR #99, opened 2023-04-10
88 scan friend qr code

### LINKED ISSUE #88: Add an option to scan friends' QR code from the app

clementcharmillot:
Add the option to scan friends' QR codes.

New button in the friendlist:
image

Clicking on it opens the following activity for result:

If the app doesn't have camera access, it is requested to the user:
image

This PR doesn't have advanced tests (actual scanning of a QR) as it is quite complicated to test the camera. If you have an idea for testing it, don't hesitate to share it. For now, the layout and the actions are tested.

The user scanned is not added to the friendlist yet as it requires another PR.

closes #88


app/src/main/java/com/epfl/drawyourpath/mainpage/fragments/FriendsFragment.kt
```diff
@@ -66,4 +71,21 @@ class FriendsFragment : Fragment(R.layout.fragment_friends) {
...
+        mainActivity.scanQRCode()
+            .thenApply {
+                // TODO: Add friend from ID "it"
+                if (it == null) {
+                    Toast.makeText(mainActivity, "Scan cancelled", Toast.LENGTH_LONG).show()
+                }
+                else {
+                    Toast.makeText(mainActivity, "Scanned $it", Toast.LENGTH_LONG).show()
+                }
+            }
```
damiengenoud, line 86: Here I don't see a way for `it` to be anything but `null.` Looking at the `scanQRCode()` function, it seems that it returns either a `CompletableFuture` with `null` as value, or completes exceptionally. Am I missing something?

app/src/main/java/com/epfl/drawyourpath/qrcode/QRScannerActivity.kt
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    private fun completeWithResult(data: String?) {
+        val resultIntent = Intent()
+        if (data != null) {
+            resultIntent.putExtra(SCANNER_ACTIVITY_RESULT_KEY, data)
+        }
+        setResult(SCANNER_ACTIVITY_RESULT_CODE, resultIntent)
+        finish()
+    }
```
damiengenoud, line 58: (Small) Detail: Some comments for these kinds of functions could help a bit with readability.

damiengenoud (approved):
Very good PR, the code is clean, clear and well modularized, thus easily understandable.
The layout is pleasant and the QR code scanner is easy/intuitive to use.

Concerning the testing, the only solution I could think of is to change the "scene" that is displayed by default on the camera with another one containing a QR code. However I don't know if it is feasible and you have certainly already  thought about this.

Otherwise nothing to say, it looks very good to me! Good job!

app/src/main/java/com/epfl/drawyourpath/mainpage/fragments/FriendsFragment.kt
clementcharmillot: The flow is a bit complex due to how Android handles messages through activities.

The flow is as followed:
1. User calls `scanQRCode()`.
2. `scanQRCode()` calls `ActivityCompat.requestPermissions()`.
3. Permission result is returned through `MainActivity.onRequestPermissionsResult()` which is part of the Android SDK.
4. `MainActivity.onRequestPermissionsResult()` calls `launchFriendQRScanner()` which starts the `QRScannerActivity`, expecting a result.
5. The user scans a QR code and the `QRScannerActivity` activity sets the result with `setResult()` and terminates to give back execution to the `MainActivity` with `finish()`.
6. `MainActivity.onActivityResult()` is called as part of the Android SDK for activity completion.
7. `MainActivity.onActivityResult()` gets the result of the scan through an intent and fires the future.

app/src/main/java/com/epfl/drawyourpath/qrcode/QRScannerActivity.kt
clementcharmillot: I've added more comments, let me know if you think I should add more.

damiengenoud: You don't need to add more it's perfectly fine like this :)

app/src/main/java/com/epfl/drawyourpath/mainpage/fragments/FriendsFragment.kt
damiengenoud: I think I understand it a bit better with your nice explaination but still, I feel like the result is contained in `qrScanResult` and not in `result`, which is what `scanQRCode()` returns. The line 86 of `MainActivity` seems to be "inverted" to me... but there is maybe still something I don't get.

clementcharmillot: We store the new future for later completion. We return the future without it being completed because we need to wait for results before firing it.

damiengenoud: Yes I think I can see the logic now, even though it's quite confusing, thank you for the explaination!

damiengenoud (approved):

PR #100, opened 2023-04-11
Add Friends List in UserModel and in the databases 

### LINKED ISSUE #81: Complete the UserModel and the Database to add the friend list

HugoHof:
This PR contains : 
- add a friendlist parameter in the userModel that consist of a list of userId of the friends of the user
- add a function to get this friendslist in the UserModel
- add a function addFriend() to userModel to add a friend to the friendliest if this friends is present on the database 
- add a function removeFriend to remove a friend from the friendslist(throw an exception if the user that we want to removed is not present on the friendliest)
- add a function to add a friend  to the friendslist of a user in the databases(throw an error if the user that we want to added is not present on the database)
- add a function to remove a user to a friendliest on the database(throw an error if the user that we want to removed is not present on the friendslist)

Closes #81

### DIFF
--- a/app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
+++ b/app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
@@ -224,7 +224,7 @@ class FireDatabase : Database() {
                 //add the user to the the friendList of the current user
                 val currentUserId = getUserId()
                 if(currentUserId==null){
-                    throw Exception("The userId can't be null !")
+                    throw Exception("Any user is logged !")
                 }else{
                     addUserIdToFriendList(currentUserId, userId).thenApply {
                         //add the currentUser to the friend list of the user with userId
@@ -239,7 +239,7 @@ class FireDatabase : Database() {
         val currentUserId = getUserId()
         if(currentUserId==null){
             val future = CompletableFuture<Unit>()
-            future.completeExceptionally(Exception("The userId can't be null !"))
+            future.completeExceptionally(Exception("Any user is logged !"))
             return future
         }else{
             //remove the userId from the friendlist of the current user
--- a/app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
+++ b/app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
@@ -215,39 +215,39 @@ class FireDatabase : Database() {
         return updateUserData(dataUpdated)
     }
 
-    override fun addUserToFriendsList(userId: String): CompletableFuture<Boolean> {
-        val future = CompletableFuture<Boolean>()
-        isUserStoredInDatabase(userId).thenAccept {
+    override fun addUserToFriendsList(userId: String): CompletableFuture<Unit> {
+        val future = CompletableFuture<Unit>()
+        return isUserStoredInDatabase(userId).thenApply {
             if(!it){
-                future.completeExceptionally(java.lang.Error("The user with $userId is not present on the database."))
+                throw Exception("The user with $userId is not present on the database.")
             }else{
-                val newFriend = HashMap<String,Any?>()
-                newFriend.put(userId, null)
-                accessUserAccountFile(userId).child(friendsListFile).updateChildren(newFriend)
-                    .addOnSuccessListener { future.complete(true) }
-                    .addOnFailureListener { future.completeExceptionally(it) }
-            }
-        }
-        return future
-    }
-
-    override fun removeUserToFriendsList(userId: String): CompletableFuture<Boolean> {
-        val future = CompletableFuture<Boolean>()
-
-        accessUserAccountFile(userId).child(friendsListFile).get()
-            .addOnSuccessListener {previousFriendsList->
-                if(!previousFriendsList.children.contains(userId as DataSnapshot)){
-                    future.completeExceptionally(java.lang.Error("The userId $userId is not in the friends list."))
+                //add the user to the the friendList of the current user
+                val currentUserId = getUserId()
+                if(currentUserId==null){
+                    throw Exception("The userId can't be null !")
                 }else{
-                    //delete the user if it is present in the friends list
-                    val updateData = previousFriendsList.children.filter { (it.key as String)!=userId }
-                    accessUserAccountFile(userId).child(userId).setValue(updateData)
-                        .addOnSuccessListener { future.complete(true) }
-                        .addOnFailureListener { future.completeExceptionally(it) }
+                    addUserIdToFriendList(currentUserId, userId).thenApply {
+                        //add the currentUser to the friend list of the user with userId
+                        addUserIdToFriendList(userId, currentUserId)
+                    }
                 }
             }
-            .addOnFailureListener { future.completeExceptionally(it) }
-        return future
+        }
+    }
+
+    override fun removeUserFromFriendlist(userId: String): CompletableFuture<Unit> {
+        val currentUserId = getUserId()
+        if(currentUserId==null){
+            val future = CompletableFuture<Unit>()
+            future.completeExceptionally(Exception("The userId can't be null !"))
+            return future
+        }else{
+            //remove the userId from the friendlist of the current user
+            return removeUserIdToFriendList(currentUserId, userId).thenApply{
+                //remove the current userId to the friendlist of the user with userId
+                removeUserIdToFriendList(userId, currentUserId)
+            }
+        }
     }
 
     /**
@@ -382,7 +382,7 @@ class FireDatabase : Database() {
      */
     private fun transformFriendsList(data: DataSnapshot?): List<String>{
         if(data == null){
-            throw java.lang.Error("The friends list is not present on the database.")
+            return emptyList()
         }
         val friendListUserIds = ArrayList<String>()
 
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
```diff
@@ -214,6 +215,41 @@ class FireDatabase : Database() {
         return updateUserData(dataUpdated)
     }
 
+    override fun addUserToFriendsList(userId: String): CompletableFuture<Boolean> {
```
clementcharmillot, line 218: We shouldn't return a boolean here. Should be `CompletableFuture<Void>` or `CompletableFuture<Unit>`.

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
```diff
@@ -214,6 +215,41 @@ class FireDatabase : Database() {
...
+        val future = CompletableFuture<Boolean>()
+        isUserStoredInDatabase(userId).thenAccept {
+            if(!it){
+                future.completeExceptionally(java.lang.Error("The user with $userId is not present on the database."))
```
clementcharmillot, line 222: Should stay with Kotlin types.
```kotlin
future.completeExceptionally(Exception("The user with $userId is not present on the database."))
```

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
```diff
@@ -214,6 +215,41 @@ class FireDatabase : Database() {
...
+            }else{
+                val newFriend = HashMap<String,Any?>()
+                newFriend.put(userId, null)
```
clementcharmillot, line 225: Does it work? I think RDB removes keys with `null` values. We should simply store it as an array.

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
```diff
@@ -214,6 +215,41 @@ class FireDatabase : Database() {
...
+                accessUserAccountFile(userId).child(friendsListFile).updateChildren(newFriend)
+                    .addOnSuccessListener { future.complete(true) }
+                    .addOnFailureListener { future.completeExceptionally(it) }
+            }
+        }
+        return future
+    }
+
+    override fun removeUserToFriendsList(userId: String): CompletableFuture<Boolean> {
```
clementcharmillot, line 234: Same remark with the boolean.
Also, I think we should name it `removeUserFromFriendList`.

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
```diff
@@ -214,6 +215,41 @@ class FireDatabase : Database() {
...
+        val future = CompletableFuture<Boolean>()
+
+        accessUserAccountFile(userId).child(friendsListFile).get()
+            .addOnSuccessListener {previousFriendsList->
+                if(!previousFriendsList.children.contains(userId as DataSnapshot)){
+                    future.completeExceptionally(java.lang.Error("The userId $userId is not in the friends list."))
+                }else{
+                    //delete the user if it is present in the friends list
+                    val updateData = previousFriendsList.children.filter { (it.key as String)!=userId }
+                    accessUserAccountFile(userId).child(userId).setValue(updateData)
```
clementcharmillot, line 244: I don't think it's supposed to access `users/$uid/$uid`.

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
```diff
@@ -214,6 +215,41 @@ class FireDatabase : Database() {
...
+        return future
+    }
+
+    override fun removeUserToFriendsList(userId: String): CompletableFuture<Boolean> {
+        val future = CompletableFuture<Boolean>()
+
+        accessUserAccountFile(userId).child(friendsListFile).get()
+            .addOnSuccessListener {previousFriendsList->
+                if(!previousFriendsList.children.contains(userId as DataSnapshot)){
+                    future.completeExceptionally(java.lang.Error("The userId $userId is not in the friends list."))
```
clementcharmillot, line 240: Same remark, we should use Kotlin types, not Java's.

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
```diff
@@ -335,6 +374,23 @@ class FireDatabase : Database() {
...
     }
+
+    /**
+     * Helper function to obtain the friends list from the database of the user
+     * @param data the data snapshot containing the user List
+     * @return a list containing the userIds of the friends
+     */
+    private fun transformFriendsList(data: DataSnapshot?): List<String>{
+        if(data == null){
+            throw java.lang.Error("The friends list is not present on the database.")
```
clementcharmillot, line 385: Use Kotlin types.
Also, if the user has no friend, we don't need to throw an error.

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
```diff
@@ -214,6 +215,41 @@ class FireDatabase : Database() {
...
+                    .addOnFailureListener { future.completeExceptionally(it) }
+            }
+        }
+        return future
+    }
+
+    override fun removeUserToFriendsList(userId: String): CompletableFuture<Boolean> {
+        val future = CompletableFuture<Boolean>()
+
+        accessUserAccountFile(userId).child(friendsListFile).get()
```
clementcharmillot, line 237: We already have the friend list from the `UserModel`, we should make the change locally and push the result. No need to download the friend list again.

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
```diff
@@ -214,6 +215,41 @@ class FireDatabase : Database() {
...
+            .addOnSuccessListener {previousFriendsList->
+                if(!previousFriendsList.children.contains(userId as DataSnapshot)){
```
clementcharmillot, line 239: Removing a friend that isn't a friend shouldn't be an error. It technically succeeded as the resulting friend list doesn't contain the target user as expected.

CristianSafta (approved):
Apart from the remarks by Clement, looks good to me. The interface is well thought out and the code is well commented. Good job.

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
HugoHof: Thanks for remarking the typo

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
HugoHof: Good remark, it will be better, you are right, since we could also delete a user without internet connection and commit the change later .

HugoHof: After thinking, we need to control this aspect. Since now we the new implementation, we add/remove a friend directly in the profile of the two users.

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
HugoHof: We need to control this aspect in the new implementation.

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
HugoHof: I have done some research before doing this friendlist. And I think it s good to do like this. You can see an example on this link : https://stackoverflow.com/questions/45780290/structuring-a-friend-list-in-a-firebase-database

HugoHof:
> Thanks for your work.
> 
> Concerning the code, I think there is a bug with how the friends are stored in the Firebase Database.

Thank for your advice. 
I have change the error type and the return type of the future. 
I have improved the functions such that when a user "1" add/remove a friend "2", when the function is call:
   - the userId of the user  "1" is add/remove from the friendliest of the user "2"
   - the userId of the user "2" is add/remove from the friendlist of the user "1"
  
I have also added some helpers functions

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
HugoHof: You have right. Thank you

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
```diff
@@ -214,6 +215,41 @@ class FireDatabase : Database() {
...
+                }
+            }
+        }
+    }
+
+    override fun removeUserFromFriendlist(userId: String): CompletableFuture<Unit> {
+        val currentUserId = getUserId()
+        if(currentUserId==null){
+            val future = CompletableFuture<Unit>()
+            future.completeExceptionally(Exception("The userId can't be null !"))
```
clementcharmillot, line 242: I think the error message is misleading. It isn't the `userId` parameter that is `null`, but the current user's which means the user isn't logged.

clementcharmillot (approved):
One last small change and we are all good for me. Nice work.

HugoHof:
> One last small change and we are all good for me. Nice work.

Thank you for your review

CristianSafta (approved):
I have implemented the add and remove friend functionality in the friend list and can confirm that addFriend() and removeFriend() from userModel work as intended.

PR #107, opened 2023-04-17
95 runs history in database

damiengenoud:
This PR adds:
- a runsHistory parameter to UserModel
- a function to add a run to the history (in UserModel) and the database
- a function to remove a run from the history (in UserModel) and the database
- a function to add a run to the database
- a function to remove a run from the database

It is worth noting that two runs cannot have the same starting time in the database. That's because these are the keys for the runs in the Firebase, and it should be impossible to start to path at the exact same time. This way, the runs are always sorted when retrieving them from the database. Similarly, the list runsHistory in UserModel is always sorted when a run is added, to be coherent with the database.

This PR also cleans the files that were modified, and moves two files which were not in the right package (Run and RunsAdapter).


app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
```diff
@@ -380,54 +431,89 @@ class FireDatabase : Database() {
...
+        if (data == null) {
+            return emptyList()
+        }
+        val runsHistory = ArrayList<Run>()
+
+        for (run in data.children) {
+            val points = ArrayList<LatLng>()
+            for (point in run.child("path").child("points").children) {
+                val lat = point.child("latitude").value as Long
+                val lon = point.child("longitude").value as Long
```
clementcharmillot, line 461: Don't we lose floating point data due to the conversion `Long` -> `Double`? Shouldn't it be read as `Double` directly?
Also, if there is no value in the database, it will crash as it calls `toDouble()` on a `null` value. We might want to handle this case.

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
```diff
@@ -330,6 +336,47 @@ class UserModel {
...
+
+    /**
+     * This function will remove a run from the history of the user in the local history and the database.
+     * @param run to be removed from the history
+     * @throws Exception if the path is not in the history
+     * @return a future that indicate if the run was correctly removed from the database
+     */
+    fun removeRunFromHistory(run: Run): CompletableFuture<Unit> {
+        if (!runsHistory.contains(run)) {
+            throw Exception("This path is not in the history !")
```
clementcharmillot, line 362: Might be cleaner to handle the exception through the future here.

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
```diff
@@ -237,19 +237,65 @@ class FireDatabase : Database() {
...
+        //check if user is correctly logged
+        val currentUserId = getUserId()
+        val future = CompletableFuture<Unit>()
+        if (currentUserId == null) {
+            future.completeExceptionally(Exception("Any user is logged !"))
+            return future
+        }
+
+        //obtain the previous history
+        accessUserAccountFile(currentUserId).child(runsHistoryFile).get()
```
clementcharmillot, line 284: Reading the entire history to remove a run seems overkill. As we have the key of the run, couldn't we just delete directly?

Also, in the worst case, we already have the history in the `UserModel`.

clementcharmillot, line 284: Reading the entire history to remove a run seems overkill. As we have the key of the run, couldn't we just delete directly?

clementcharmillot (changes requested):
Nice work, the code is really clean.

Just a few remarks on the logic of the code and some edge cases and we should be all good.

Also, I think it's a good idea to run the formatter on the repo. We might want to do it in its own PR at the end of the sprint to prevent conflicting with other PRs. What do you think? As it's already done, I see no problem to let it for this PR.

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
damiengenoud: You are absolutely right, they should be read as Double, good catch. 
With this change however we have no call to toDouble() anymore. Do you still think I should check for null ? If I do so I should do it for every value I read, even though it should be impossible to store runs with points without latitude or longitude... What do you think ?

app/src/main/java/com/epfl/drawyourpath/userProfile/UserModel.kt
damiengenoud: I agree, I'm changing this.

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
damiengenoud: There is indeed a much cleaner and direct way to do it, thank you for pointing this out. I'm changing this code.

damiengenoud:
> Nice work, the code is really clean.
> 
> Just a few remarks on the logic of the code and some edge cases and we should be all good.
> 
> Also, I think it's a good idea to run the formatter on the repo. We might want to do it in its own PR at the end of the sprint to prevent conflicting with other PRs. What do you think? As it's already done, I see no problem to let it for this PR.

Thank you for the review, you pointed very important issues!

Concerning the formatter, I thought everybody used it but it's not the case. I agree that we shouldn't run it on the whole repo without talking about it with the team, but I think it's important that everyone runs it before pushing. In this PR I did not pay enough attention on which files I accidentally formatted that's why there is some "unwanted" changes. 
I do think however that the files that are actually modified in a PR should be formatted without hesitation.

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
clementcharmillot: I think we should still check for `null` in case the data is badly set in the database. Crashing the app with a `null pointer exception` might quickly become hard to trackdown for this case.

On another note, do you know how the type is handled by Firebase for a JSON `Number`? I've seen your test data is integers so I'm wondering if it will not return a `Long` in this case and a `Double` when it has decimals? If that's the case, we might need to handle both...

damiengenoud: Alright I will check for null points then, however if the start or end time is not correct I guess it will crash anyway. 
As I said I don't see how points could be badly formatted as they are basic LatLong objects when stored, so this should not be an issue and such runs should never exist.

I don't know how it is stored precisely, but will check with real values that are not integers and let you know. Taking my tests into consideration is not relevant though, as they use the mock database.

damiengenoud: @clementcharmillot  Sorry actually you are right I had to do a small change for it to work correctly with both types. Again good catch thank you!

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
```diff
@@ -380,54 +424,93 @@ class FireDatabase : Database() {
...
+        val runsHistory = ArrayList<Run>()
+
+        for (run in data.children) {
+            val points = ArrayList<LatLng>()
+            for (point in run.child("path").child("points").children) {
+                val lat = point.child("latitude").getValue(Double::class.java)
+                val lon = point.child("longitude").getValue(Double::class.java)
+                if (lat != null && lon != null) {
+                    points.add(LatLng(lat, lon))
+                }
```
clementcharmillot, line 457: As it is not normal behavior to have a null point, it might be worth to log the problem.

```kotlin
android.util.Log.w("DYP", "A run had an invalid coordinate")
```

or something similar.

clementcharmillot (approved):
Just one last small thing and we can merge, great job.

app/src/main/java/com/epfl/drawyourpath/database/FireDatabase.kt
damiengenoud: Good idea! I log the problem if the point is invalid in the following way:
`android.util.Log.w(FireDatabase::class.java.name, "A point of a run has invalid coordinates => ignoring the point")`
The tag is the class name to identify where the issue comes from easily.

PR #118, opened 2023-04-21
30 current map location

### LINKED ISSUE #30: Create a map activity showing the current location

Gafragni:
This is a draft PR for getting the current location.
Getting the permisions works and the location can be obtained but it does not update the map automatically.

closes #30


clementcharmillot:
Code discussed:
```kotlin
// Request location updates every 10 seconds
val locationRequest = LocationRequest.create()
        .setInterval(10000)
        .setFastestInterval(5000)
        .setPriority(LocationRequest.PRIORITY_HIGH_ACCURACY)

fusedLocationClient.requestLocationUpdates(locationRequest, object : LocationCallback() {
    override fun onLocationResult(locationResult: LocationResult?) {
        locationResult ?: return
        for (location in locationResult.locations) {
            // Move the camera to the user's updated location
            val latLng = LatLng(location.latitude, location.longitude)
            mMap.animateCamera(CameraUpdateFactory.newLatLngZoom(latLng, 16f))
        }
    }
}, Looper.getMainLooper())
```

app/src/main/java/com/epfl/drawyourpath/mainpage/fragments/DrawFragment.kt
```diff
@@ -1,34 +1,148 @@
...
+        }
+        try {
+            if (locationPermissionGranted) {
+                map?.isMyLocationEnabled = true
+                map?.uiSettings?.isMyLocationButtonEnabled = true
+            } else {
+                map?.isMyLocationEnabled = false
+                map?.uiSettings?.isMyLocationButtonEnabled = false
+                lastKnownLocation = null
+                getLocationPermission()
```
clementcharmillot, line 96: I think we have an indirect infinite loop `getLocationPermission()` -> `onRequestPermissionsResult()` -> `updateLocationUI()` -> `getLocationPermission()` -> ...

From my understanding, permission request is asked directly on creation so we don't need to ask it again here.

app/src/main/java/com/epfl/drawyourpath/mainpage/fragments/DrawFragment.kt
```diff
@@ -1,34 +1,148 @@
...
+                    .setInterval(LOCATION_REQUEST_INTERVAL)
+                    .setFastestInterval(LOCATION_REQUEST_MIN_INTERVAL)
+                    .setPriority(LocationRequest.PRIORITY_HIGH_ACCURACY)
+
+                fusedLocationProviderClient.requestLocationUpdates(locationRequest, object : LocationCallback() {
+                    override fun onLocationResult(locationResult: LocationResult?) {
+                        if (locationResult != null) {
+                            for (location in locationResult.locations) {
+                                map?.animateCamera(CameraUpdateFactory.newLatLng(LatLng(location.latitude, location.longitude)))
+                            }
```
clementcharmillot, line 118: The last element of `locationResult.locations` is the most recent location received for the user. We probably only want to animate the camera to this location.

clementcharmillot (changes requested):
Nice PR, the code is clean.

I have a few comments about the code.

On another note. I think we want to extract the getting location part from the fragment directly. As we will need to track the user location and store it to track his path in real time. The map should then use this data. However, this should probably be done as another task in the next sprint.

app/src/main/java/com/epfl/drawyourpath/mainpage/fragments/DrawFragment.kt
```diff
@@ -1,34 +1,148 @@
...
+                // If request is cancelled, the result arrays are empty.
+                if (grantResults.isNotEmpty() &&
+                    grantResults[0] == PackageManager.PERMISSION_GRANTED) {
+                    locationPermissionGranted = true
+                }
+            }
+            else -> super.onRequestPermissionsResult(requestCode, permissions, grantResults)
+        }
+        updateLocationUI()
+    }
```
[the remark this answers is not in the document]
Gafragni: Right, it was implemented as if there was no other code than requesting this location permission.
Your implementation looks perfect.

After some debugging I realised this function was actually never called, when the permission results came it was the MainActivity's onRequestPermissionsResult that was called. Fixed by correcting getLocationPermission()

app/src/main/java/com/epfl/drawyourpath/mainpage/fragments/DrawFragment.kt
Gafragni: Yes we do have an infinite loop, at first I thought it was intended but after modifying the code, it is clearly not.
Simple fix, remove the getLocationPermission() call in updateLocationUI()

app/src/main/java/com/epfl/drawyourpath/mainpage/fragments/DrawFragment.kt
Gafragni: Yes, we can use locationResult.lastLocation instead. 
Overall this leads to fewer camera movements.

app/src/main/java/com/epfl/drawyourpath/mainpage/fragments/DrawFragment.kt
```diff
@@ -1,34 +1,148 @@
...
-        map.addMarker(
-            MarkerOptions()
-                .position(coordinates)
-                .title("EPFL"),
-        )
-        map.moveCamera(CameraUpdateFactory.newLatLng(coordinates))
-        map.animateCamera(CameraUpdateFactory.newLatLngZoom(coordinates, 15F))
+        this.map = map
+        map?.moveCamera(CameraUpdateFactory.newLatLngZoom(DEFAULT_LOCATION, DEFAULT_ZOOM))
+        getLocationPermission()
```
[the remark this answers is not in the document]
Gafragni: Yes we shoud, now it asks the permissions as the fragment is created.

clementcharmillot (approved):
Thanks for the changes. Looks good to me now, we can merge.

PR #131, opened 2023-04-21
127 git hook autoformat

### LINKED ISSUE #127: Setup Git Hook to check for formatting before commits

clementcharmillot:
Adds a Github Action that automatically formats the code on the branch after a push and pushes the result on the same branch.

If the code is compliant to the coding style, no action is taken.

You can see the result of the action on commit a1402ae17e12a3fe4b319f4bcbb44e62cc9af605.

To isolate changed files, see commit ccf98ad18e837eb44c8f6735b7723aa141e0c616.

Question: Should it run only on main after merge?

closes #127


app/src/androidTest/java/com/epfl/drawyourpath/mainpage/fragments/ChallengeFragmentTest.kt
```diff
@@ -222,25 +221,27 @@ class ChallengeFragmentTest {
...
-                    0, replaceTextOnViewChild("", R.id.goal_display_edit_text)
+                    0,
+                    replaceTextOnViewChild("", R.id.goal_display_edit_text),
                 ),
                 actionOnItemAtPosition<RecyclerView.ViewHolder>(
-                    0, pressImeActionButtonOnViewChild(R.id.goal_display_edit_text)
-                )
+                    0,
+                    pressImeActionButtonOnViewChild(R.id.goal_display_edit_text),
+                ),
```
damiengenoud, line 237: Remark: there are these types of changes that I'm not sure are really useful but I guess it is fine to still have them.

damiengenoud (approved):
Thank you for adding this so quickly! I think it is fine if we only run it when we merge on main. By the way it seems that the formatting is not the same as the default one of Android Studio, but at least with that we are pulling cleaner code.

clementcharmillot:
Thanks for the review. The formatter has some configuration that we may play with.

I've changed the execution policy to only run on main. I've setup the action to bypass main branch protection but couldn't test it on my own branch as it isn't enabled. Should we still commit and potentially create another PR to solve any issue the action might face due to branch protection?

damiengenoud:
If you think that it's unlikely that we will have problems then you can merge for me. As we are in bug fixing phase right now it should not be a problem to open a PR if problems arise.

PR #134, opened 2023-04-28
124 added chat and message Kotlin classes 

### LINKED ISSUE #124: Create a message class that represents a message along with usefull methods.

CristianSafta:
**Implement Chat and Message classes with MessageContent**

Description:

This pull request implements the Chat and Message classes to model a chat feature for users in the running app. The Message class includes a sealed MessageContent class to represent three types of message content: Text, Picture, and RunPath.

Changes made:

Added a sealed MessageContent class with three subclasses:

-Text: Represents a text message with a text property.
-Picture: Represents an image message with an image property (Bitmap).
-RunPath: Represents a run path message with a run property (Run object).
Created a Message class to model a chat message with the following properties:

-id: The unique identifier for the message.
-senderId: The unique identifier for the sender of the message.
-content: The MessageContent object representing the content of the message.
-timestamp: The timestamp of the message (in milliseconds since epoch).
The Message class includes factory methods for creating each type of message (Text, Picture, and RunPath).
Implemented a Chat class to model a chat conversation with the following methods:

-addMessage(): Adds a message to the chat.
-getMessages(): Returns a list of all messages in the chat.
-getMessageById(): Retrieves a message by its unique identifier.
-removeMessageById(): Removes a message by its unique identifier.
Updated the MessageContent.Text class with an updateTextContent() method for updating the text content of a message.

Created unit tests for the Chat and Message classes in the ChatTest class to ensure the proper functionality of the implemented features.

Closes #124

### DIFF
--- a/app/src/main/java/com/epfl/drawyourpath/chat/Chat.kt
+++ b/app/src/main/java/com/epfl/drawyourpath/chat/Chat.kt
@@ -1,0 +1,147 @@
+package com.epfl.drawyourpath.chat
+
+
+import android.graphics.Bitmap
+import com.epfl.drawyourpath.path.Run
+
+/**
+ * A sealed class representing the content of a chat message.
+ * There are three types of message content: Text, Picture, and RunPath.
+ */
+sealed class MessageContent {
+    /**
+     * Represents a text message.
+     *
+     * @property text The text content of the message.
+     */
+    data class Text(var text: String) : MessageContent() {
+        /**
+         * Update the content of a text message
+         *
+         * @param newText The new text content for the message
+         */
+        fun updateTextContent(newText: String) {
+            text = newText
+        }
+    }
+
+    /**
+     * Represents an image message.
+     *
+     * @property image The Bitmap image of the message.
+     */
+    data class Picture(val image: Bitmap) : MessageContent()
+
+    /**
+     * Represents a run path message.
+     *
+     * @property run The Run object containing the run path data.
+     */
+    data class RunPath(val run: Run) : MessageContent()
+}
+
+/**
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/epfl/drawyourpath/chat/Chat.kt
```diff
@@ -0,0 +1,147 @@
...
+}
+
+/**
+ * A class representing a chat message.
+ *
+ * @property id The unique identifier for the message.
+ * @property senderId The unique identifier for the sender of the message.
+ * @property content The MessageContent object representing the content of the message.
+ * @property timestamp The timestamp of the message (in milliseconds since epoch).
+ */
```
HugoHof, line 50: Maybe we can use the timestamp of the message as the message id since it is theoretically impossible that the same user send a message at the same time. With that, it will take less memory on the database.

HugoHof (approved):
This is a good PR. This class will be very useless to make the link between the database and the UI of the chat. The code is well commented and well structured. 
Suggestion: we can replace the message ID by the timestamp directly to take less memory on the database (since it is theoretically impossible that two message are send exactly at the same time in milliseconds). 
Don't hesitate if you have any question.

app/src/main/java/com/epfl/drawyourpath/chat/Chat.kt
CristianSafta: Good suggestion. I have implemented it. Although in this case, it would be good to assume that timestamps will only be unique in the context of a chat, otherwise it would still be possible(although rare) that two messages have the same id.

PR #137, opened 2023-04-29
136-fix-formatter

### LINKED ISSUE #136: fix formatter

clementcharmillot:
Changes the way the formatter runs:

The formatter now runs on each branch that is not main after a commit.

closes #136


damiengenoud (approved):
Thank you for the quick and necessary change!

PR #142, opened 2023-05-03
125 create UI to draw a path

### LINKED ISSUE #125: create-UI-to-draw-a-path

HugoHof:
This PR contains a UI to draw a path. We can lunch this activity on the main menu by clicking on a start drawing button. Then a countdown will be lunch. 
Capture d’écran 2023-05-03 à 22 24 32

After that, a fragment containing the position of the user(with the current path draw on the map) will be displayed on the map, some current information on the sport performance of the user(time, speed, distance) and a button to pause the drawing. 
Capture d’écran 2023-05-03 à 22 23 49

By clicking on pause, the map will displayed the current path draw by the user(global view of the path centered on middle of the path), sort information and a button to resume the drawing and another button to stop the drains.

Capture d’écran 2023-05-03 à 22 24 08

When we stop the drawing, a view will show the path down by the user and the sport performance of the user(average time taken by the user to do 1km, average speed, distance, time, calories burned) and a button to go back to the menu.
Capture d’écran 2023-05-03 à 22 24 22

Close #125 

Remark : the path is not displayed on the pictures below the map, but all the functions to draw the path are present(and tested)


app/src/main/java/com/epfl/drawyourpath/pathDrawing/PathDrawingCountDownFragment.kt
```diff
@@ -0,0 +1,48 @@
...
+            override fun onTick(millisUntilFinished: Long) {
+                val seconds = (millisUntilFinished) / 1000
+                countDownText.text = if (seconds == 0L) "GO !" else seconds.toString()
+            }
+
+            override fun onFinish() {
+                if (activity != null) {
+                    displayedMainDrawFragment()
+                }
+            }
```
HoLuDyEr, line 34: This could be inside a function instead of having one big function for this fragment.
something like `private fun createCountDown(seconds: Double): CountDownTimer`

app/src/androidTest/java/com/epfl/drawyourpath/pathDrawing/PathDrawingCountDownFragmentTest.kt
```diff
@@ -0,0 +1,70 @@
...
+        )
+        val t: ActivityScenario<PathDrawingActivity> = ActivityScenario.launch(intent)
+
+        Espresso.onView(ViewMatchers.withId(R.id.path_drawing_countdown_fragment))
+            .check(ViewAssertions.matches(ViewMatchers.isDisplayed()))
+        Thread.sleep(4000)
+        Espresso.onView(ViewMatchers.withId(R.id.path_drawing_main_fragment))
+            .check(ViewAssertions.matches(ViewMatchers.isDisplayed()))
+        t.close()
+    }
```
HoLuDyEr, line 69: It could be good to have this inside the other test too because the thread.sleep takes a long time so that we wait only once for the timeout and not twice.

HoLuDyEr: or maybe add the timeout time as parameters and for this test setting it to a lower timeout.

app/src/androidTest/java/com/epfl/drawyourpath/pathDrawing/PathDrawingEndFragmentTest.kt
```diff
@@ -0,0 +1,108 @@
...
+     */
+    @Test
+    fun checkThatClickingOnBackMenuButtonShowMainAppActivity() {
+        val intent = Intent(
+            ApplicationProvider.getApplicationContext(),
+            PathDrawingActivity::class.java,
+        )
+        val t: ActivityScenario<PathDrawingActivity> = ActivityScenario.launch(intent)
+        // wait that the countdown passed
+        Thread.sleep(4100)
```
HoLuDyEr, line 51: if we can change the countdown time it would make the test finish faster

HoLuDyEr (changes requested):
The code is well written and documented, there are only a few small remarks.
I have two general remarks for the tests. The first is a small detail, the Espresso, ViewMatchers,... can be omitted and make the test a bit less verbose.
The second one is for the timeouts, it would be good to either having a parameter to skip the timeout or having less time in the timeout.

HugoHof:
> The code is well written and documented, there are only a few small remarks.
> I have two general remarks for the tests. The first is a small detail, the Espresso, ViewMatchers,... can be omitted and make the test a bit less verbose.
> The second one is for the timeouts, it would be good to either having a parameter to skip the timeout or having less time in the timeout.

Thank you for your review. Thank you for your advice, I have apply them as much as possible.

app/src/main/java/com/epfl/drawyourpath/map/MapFragment.kt
```diff
@@ -68,22 +68,31 @@ class MapFragment(private val focusedOnPosition: Boolean = true, private val pat
...
             drawPathOnMap(map, path!!)
         }
     }
 
+    /**
+     * Function used to move the camera at a given location
+     * @param location where we want to move the camera
+     * @param zoom apply to the camera
+     */
+    private fun moveCameraToPosition(location: LatLng, zoom: Float) {
```
HoLuDyEr, line 84: This function can be used inside the function `getDeviceLocation()`

HoLuDyEr (approved):
Thank you for the changes, there is still a minor detail but I will already approve as it is a quick change.

app/src/main/java/com/epfl/drawyourpath/map/MapFragment.kt
HugoHof: It's done now thank you.

PR #143, opened 2023-05-03
120 basic db operations for tournaments

### LINKED ISSUE #96: Link tournaments with the database

### LINKED ISSUE #120: Modify tournaments to fit in database

damiengenoud:
This PR adds some basic functions to create tournaments and update its users in the database.
The main features are:
- Adding a file called `tournaments` at the database root, which stores information about all tournaments.
- Adding a subfile to `users` also called `tournaments` which stores the id of the tournaments in which the user participates.
- Adding two functions to add  (`addTournament()`) and remove (`removeTournament()`) tournaments from the database.  Note that using the remove function will remove the tournament from all participants' `tournaments` subfile.
- Adding two function to add a user (`addUserToTournament()`) and remove a user (`removeUserFromTournament()`) from a tournament.
- Modifying the tournament class so that it contains more relevant information on the tournament (like the creator, the participants, ...)
- Modifying the tournament creation fragment so that it stores the tournament in the database (even though we cannot yet retrieve the tournament). The tournament stored will not have any participant or post.

Notes:
- In case of failure when creating a tournament, the exception is handled and displays a toast notifying the user that the operation failed. I have however not been able to test that the toast is indeed displayed, even after a lot of research.
- If the user is offline when creating the tournament, Firebase caches it and still stores it when the user gets back online.
- It is not yet possible to retrieve the tournaments from the database, but it should be done in a future sprint.

closes #120, closes #96


app/src/main/java/com/epfl/drawyourpath/community/TournamentCreationFragment.kt
```diff
@@ -54,10 +89,99 @@ class TournamentCreationFragment : Fragment(R.layout.fragment_tournament_creatio
...
+            // Display toasts when complete. Note that it does not complete while offline (but the tournament
+            // is stored when the user gets back online)
+            if (exception != null) {
+                Toast.makeText(
+                    mActivity,
+                    "Operation failed: ${exception.message}",
+                    Toast.LENGTH_LONG,
+                ).show()
+            } else {
+                Toast.makeText(mActivity, "Tournament created!", Toast.LENGTH_LONG).show()
```
clementcharmillot, line 184: Could redirect to the tournament list or the main activity maybe?

app/src/main/java/com/epfl/drawyourpath/database/FirebaseDatabase.kt
```diff
@@ -463,6 +488,86 @@ class FirebaseDatabase : Database() {
         return Utils.failedFuture(Error("Not implemented"))
     }
 
+    override fun getTournamentUID(): String? {
+        return database.child(FirebaseKeys.TOURNAMENTS_ROOT).push().key
```
clementcharmillot, line 492: Doesn't it make a server request? . If we want to generate a collision resistant id locally, we should use 
```kotlin
UUID.randomUUID().toString()
```

app/src/main/java/com/epfl/drawyourpath/database/FirebaseDatabase.kt
```diff
@@ -463,6 +488,86 @@ class FirebaseDatabase : Database() {
...
+                // 2. remove the tournament from the tournaments file
+                changes.put("${FirebaseKeys.TOURNAMENTS_ROOT}/$tournamentId", null)
+                // do the changes
+                database.updateChildren(changes).addOnSuccessListener { future.complete(Unit) }
+                    .addOnFailureListener { future.completeExceptionally(it) }
+            }
+            .addOnFailureListener { future.completeExceptionally(it) }
+
+        return future
+    }
```
clementcharmillot, line 526: I don't think we can make this code work with security rules.
Also, if we have let's say a million participant registered to a tournament, a client will make a million request emptying his mobile data. It doesn't really scale.

I think we could keep it like that for this sprint, but we should only mark a tournament as deleted and keep it in the db. We probably are going to want to show a list of past tournaments. What do you think?

app/src/main/java/com/epfl/drawyourpath/database/FirebaseDatabase.kt
```diff
@@ -463,6 +488,86 @@ class FirebaseDatabase : Database() {
...
+
+    override fun addUserToTournament(
+        userId: String,
+        tournamentId: String,
+    ): CompletableFuture<Unit> {
+        val future = CompletableFuture<Unit>()
+        // check that the userId and tournamentId exist
+        isUserInDatabase(userId).thenCombine(isTournamentInDatabase(tournamentId)) { userExists, tournamentExists ->
+            run {
```
clementcharmillot, line 535: Why use `run` here?

Apart from that, this method is really well implemented with a good use of the `Future` API. Also nice to do the two writes in one go 👍.

app/src/main/java/com/epfl/drawyourpath/mainpage/fragments/CommunityFragment.kt
```diff
@@ -145,7 +152,13 @@ class CommunityFragment : Fragment(R.layout.fragment_community) {
...
             .setIcon(R.drawable.ic_add)
             .setOnMenuItemClickListener {
-                replaceFragment<TournamentCreationFragment>()
+                // send args for mock to tournament creation fragment
+                val db = activity?.intent?.getStringExtra("Database")
+                val auth = activity?.intent?.getStringExtra("Auth")
+                val fragmentArgs = Bundle()
+                if (db != null) fragmentArgs.putBoolean(db, true)
+                if (auth != null) fragmentArgs.putBoolean(auth, true)
+                replaceFragment<TournamentCreationFragment>(fragmentArgs)
```
clementcharmillot, line 161: What is this code for? It's really hard to understand.

After reading the tests I understood, but isn't it way too complicated? Why not just pass a boolean with a key? 

```
USE_MOCK_IMPLEMENTATION to true / false
USE_MOCK_FAILING_IMPLEMENTATION to true / false
```

clementcharmillot (changes requested):
Nice work, most of the code is clean.

I have just a few small remarks that shouldn't take long to fix, and a discussion about deleting tournaments.

app/src/main/java/com/epfl/drawyourpath/community/TournamentCreationFragment.kt
```diff
@@ -30,6 +49,22 @@ class TournamentCreationFragment : Fragment(R.layout.fragment_tournament_creatio
...
+        val startDate: LocalDateTime,
+        val endDate: LocalDateTime,
+        val visibility: Tournament.Visibility,
+    )
+
+    override fun onAttach(context: Context) {
+        super.onAttach(context)
+        if (context is Activity) {
+            mActivity = context
+        }
```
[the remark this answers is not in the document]
damiengenoud: Well when I tried to make the toasts appear it didn't work through `activity,` but I must have made something wrong because you're right, it doesn't change anything. Changing it back to `activity`. Good catch!

app/src/main/java/com/epfl/drawyourpath/community/TournamentCreationFragment.kt
```diff
@@ -54,10 +89,99 @@ class TournamentCreationFragment : Fragment(R.layout.fragment_tournament_creatio
...
+    }
+
+    /**
+     * Helper function that returns a database based on the arguments passed to
+     */
+    private fun getDatabase(): Database {
+        return if (arguments?.getBoolean("USE_WORKING_MOCK_DB", false) == true) {
+            MockDatabase()
+        } else if (arguments?.getBoolean("USE_FAILING_MOCK_DB", false) == true) {
```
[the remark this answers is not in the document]
damiengenoud: Where would you put these variables? I'm putting them in the companion object of the tournament for now, let me know what you think.

app/src/main/java/com/epfl/drawyourpath/community/TournamentCreationFragment.kt
```diff
@@ -54,10 +89,99 @@ class TournamentCreationFragment : Fragment(R.layout.fragment_tournament_creatio
...
+            MockNonWorkingDatabase()
+        } else {
+            FirebaseDatabase()
+        }
+    }
+
+    private fun getAuth(): Auth {
+        return if (arguments?.getBoolean("USE_WORKING_MOCK_AUTH", false) == true) {
+            MockAuth(forceSigned = true)
+        } else if (arguments?.getBoolean("USE_FAILING_MOCK_AUTH", false) == true) {
```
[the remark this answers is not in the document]
damiengenoud: Also putting them in the companion object for now.

app/src/androidTest/java/com/epfl/drawyourpath/community/TournamentCreationFragmentTest.kt
```diff
@@ -25,22 +30,49 @@ import java.time.LocalTime
...
-    @Before
-    fun launchFragment() {
+    private fun launchFragmentFromMainActivity(
+        workingDB: Boolean,
+        workingAuth: Boolean,
+    ): ActivityScenario<MainActivity> {
+        Intents.init()
+
+        val argDB = if (workingDB) "USE_WORKING_MOCK_DB" else "USE_FAILING_MOCK_DB"
+        val argAuth = if (workingAuth) "USE_WORKING_MOCK_AUTH" else "USE_FAILING_MOCK_AUTH"
```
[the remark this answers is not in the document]
damiengenoud: Changed!

app/src/main/java/com/epfl/drawyourpath/community/TournamentCreationFragment.kt
```diff
@@ -54,10 +89,99 @@ class TournamentCreationFragment : Fragment(R.layout.fragment_tournament_creatio
     private fun createCreateButton(view: View) {
         val createButton = view.findViewById<Button>(R.id.tournament_creation_create_button)
         createButton.setOnClickListener {
```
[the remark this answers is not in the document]
damiengenoud: Yes it would be much cleaner this way, I'm changing this. Thank you for pointing it out. I tried not touching too much of what was already done but some part of the logic and functions naming are debatable to say the least...

app/src/main/java/com/epfl/drawyourpath/community/TournamentCreationFragment.kt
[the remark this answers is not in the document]
clementcharmillot: As it's really linked to the fragment, I would declare them as constant variable in `TournamentCreationFragment.kt`. Putting them in the companion object of the fragment is also clean.

app/src/main/java/com/epfl/drawyourpath/community/TournamentCreationFragment.kt
damiengenoud: This is part of what happens when the future returned by the database is completed, which can take quite some time. By this time, the last instruction of the `OnCreateTournamentClicked()` which returns to the Community fragment has usually already been called. So it's already returning to the main activity. It works like this so that the user leaves the tournament creation fragment even if the database takes some time to respond.

clementcharmillot: I see, but what if an error occurs? Does the user lose all the data he has set?

app/src/main/java/com/epfl/drawyourpath/database/FirebaseDatabase.kt
damiengenoud: It works perfectly offline so I don't think it does. Morevoer, every examples I've seen use push() (with no argument) to get a unique ID. If in the future we realize that it actually creates some problems we could easily change it though.

app/src/main/java/com/epfl/drawyourpath/database/FirebaseDatabase.kt
damiengenoud: I forgot to remove it, good catch. 
And thank you, I've had a hard time figuring these futures out!

app/src/main/java/com/epfl/drawyourpath/mainpage/fragments/CommunityFragment.kt
damiengenoud: Well this code is the best solution I could find to pass the arguments to the tournament creation fragment, because we have 5 possible cases:
- Firebase DB + Auth
- Mock working DB + Mock working auth
- Mock working DB + Mock failing auth
- Mock failing DB + Mock working auth 
- Mock failing DB + Mock failing auth
- 
I'm adding a comment to say that it only matters when testing, otherwise `db`  and `auth` are null anyway.

app/src/main/java/com/epfl/drawyourpath/database/FirebaseDatabase.kt
damiengenoud: I will be honest I have no idea about the security rules, but I guess that a user deleting a tournament should not change the user file of another user? If it's true then this function could be reserved to "admins"... I don't really know.

Indeed it is a problem that it doesn't scale well and I did not think about it. However I would argue that it is still useful to have such a function to be able to really remove a tournament. Maybe, as I said above, we could only use it in really special situations, when we don't have much choice (but only as admins). 

We certainly should not let users delete tournaments with this function all the times if there are large numbers of participants, and we should be careful when deleting weekly tournament. 

I can see two solutions for now:
- limit the number of participants in tournaments that are not weekly (but requires to distinguish the weekly tournaments).
- "never" delete tournaments as you said, except if the tournament has not started for example.

Unfortunately I don't have time to change all of this for this sprint, but we should talk about that with the whole team on a meeting, to see what should be done.

I hope everything I said makes sense, let me know what you think.

damiengenoud:
@clementcharmillot 
Thank you for the detailed review, I've made the changes you requested. Let me know if anything else should be modified or is still unclear.

app/src/main/java/com/epfl/drawyourpath/community/TournamentCreationFragment.kt
damiengenoud: Yes, but it's not a big deal I think. And it seems that errors very rarely happen with Firebase it seems, it works even if the user is offline.

app/src/androidTest/java/com/epfl/drawyourpath/database/MockDatabaseTest.kt
```diff
@@ -55,6 +56,26 @@ class MockDatabaseTest {
         assertEquals(test, false)
     }
 
+    /**
+     * Test if a tournamentId present in the database is returned as present
+     */
+    @Test
```
clementcharmillot, line 62: We might add one test for the failing database as well for coverage. Maybe also for `delete`, `add`, ...  operations on tournament.

app/src/main/java/com/epfl/drawyourpath/database/FirebaseDatabase.kt
clementcharmillot: If it doesn't then your solution is even better. We can keep it like that.

app/src/main/java/com/epfl/drawyourpath/database/FirebaseDatabase.kt
clementcharmillot: I think the second option is best. We can talk about it in the next meeting and plan it for the next sprint.

clementcharmillot (changes requested):
Thanks for the changes, a small note on tests and we can merge.

Edit: I wanted to approve the PR but misclicked.

app/src/androidTest/java/com/epfl/drawyourpath/database/MockDatabaseTest.kt
damiengenoud: If you think about the `MockNonWorkingDatabase()` then the tests are in the related test class `MockNonWorkingDatabaseTest`, are you talking about other failing scenarios ?

clementcharmillot: Yes that's what I meant. It's all good then.

clementcharmillot (approved):
Thanks for the changes, we can merge.

PR #145, opened 2023-05-03
139 fix user model cached

### LINKED ISSUE #139: fix userModelCached

HoLuDyEr:
This PR fixes and cleans the UserModelCached along with its tests.
It also deletes the old UserModel.

closes #139

### DIFF
--- a/app/src/main/java/com/epfl/Utils/drawyourpath/Utils.kt
+++ b/app/src/main/java/com/epfl/Utils/drawyourpath/Utils.kt
@@ -9,9 +9,10 @@ import com.epfl.drawyourpath.R
 import com.epfl.drawyourpath.database.UserGoals
 import com.google.android.gms.maps.model.LatLng
 import java.io.ByteArrayOutputStream
-import java.time.LocalDate
+import java.time.*
 import java.util.*
 import java.util.concurrent.CompletableFuture
+import kotlin.math.roundToInt
 
 /**
  * A class containing various utility functions.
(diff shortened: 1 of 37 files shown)


app/src/main/java/com/epfl/Utils/drawyourpath/Utils.kt
```diff
@@ -30,23 +35,125 @@ object Utils {
...
     /**
      * Decodes the photo from base64 string to bitmap format and return null if the dataSnapShot is null
      * @param photoStr photo encoded
      * @return the photo in bitmap format, and null if no photo is stored on the database
      */
     fun decodePhoto(photoStr: String): Bitmap {
-        val tabByte = Base64.getDecoder().decode(photoStr as String)
+        val tabByte = Base64.getDecoder().decode(photoStr)
         return BitmapFactory.decodeByteArray(tabByte, 0, tabByte.size)
     }
```
clementcharmillot, line 68: There is code duplication. We should add 
```kotlin
fun decodePhoto(photo: ByteArray): Bitmap {
    return BitmapFactory.decodeByteArray(photo, 0, tabByte.size)
}
```
and use it in `decodePhoto(photoStr: String)` and `decodePhotoOrGetDefault(photo: ByteArray?, res: Resources): Bitmap`.

Also, `decodePhoto(photoStr: String)` should use `decodeStringAsByteArray()`.

app/src/main/java/com/epfl/drawyourpath/userProfile/UserProfile.kt
```diff
@@ -0,0 +1,110 @@
...
+        Goals(userEntity.goalAndAchievements),
+        { Utils.decodePhotoOrGetDefault(userEntity.profilePhoto, it) },
+    )
+
+    constructor(userData: UserData) : this(
+        userData.userId ?: "invalid id",
+        userData.username ?: "",
+        userData.email ?: "",
+        userData.firstname ?: "",
+        userData.surname ?: "",
```
clementcharmillot, line 76: We could use sentinels as you did for the id here to easily track a missing field.
```suggestion
        userData.username ?: "USERNAME_INVALID",
        userData.email ?: "EMAIL_INVALID",
        userData.firstname ?: "FIRSTNAME_INVALID",
        userData.surname ?: "SURNAME_INVALID",

```

app/src/main/java/com/epfl/drawyourpath/userProfile/cache/UserModelCached.kt
```diff
@@ -197,66 +189,28 @@ class UserModelCached(application: Application) : AndroidViewModel(application)
...
     /**
-     * * // TODO: Code duplication for with updateActivityTimeGoal etc. refactor.
-     * Use this function to modify the daily distance goal of the user
+     * Use this function to modify the daily goals of the user
      * @param distanceGoal new daily distance goal
+     * @param activityTimeGoal new daily time goal
+     * @param pathsGoal new daily paths goal
      */
-    fun updateDistanceGoal(distanceGoal: Double): CompletableFuture<Unit> {
+    fun updateGoals(distanceGoal: Double? = null, activityTimeGoal: Double? = null, pathsGoal: Int? = null): CompletableFuture<Unit> {
```
clementcharmillot, line 197: Should probably replace the argument list with the `UserGoals` data class to be more flexible.

app/src/main/java/com/epfl/drawyourpath/userProfile/dailygoal/DailyGoalDao.kt
```diff
@@ -101,21 +101,33 @@ interface DailyGoalDao {
...
      * @param distanceGoal the new distance goal of the user
+     * @param timeGoal the new time goal of the user
+     * @param pathsGoal the new paths goal of the user
      * @return the new daily goal
      */
     @Transaction
-    fun updateDistanceGoal(userId: String, date: Long, distanceGoal: Double): DailyGoalEntity {
-        updateDistanceGoalUser(userId, distanceGoal)
-        insertIfDailyGoalUpdateFailed(userId, date, updateDistanceGoalDailyGoal(userId, date, distanceGoal))
+    fun updateGoals(userId: String, date: Long, distanceGoal: Double? = null, timeGoal: Double? = null, pathsGoal: Int? = null): DailyGoalEntity {
```
clementcharmillot, line 113: Same remark here, we can use `UserGoals` to avoid a large parameter list. Also, makes it easier to add a goal without changing all the function definitions.

clementcharmillot (changes requested):
Great work, the code base is way cleaner that way, thanks for the PR.

I just have a few remarks and we can merge for me.

app/src/main/java/com/epfl/Utils/drawyourpath/Utils.kt
HoLuDyEr: Thank you for pointing this out, I also changed the return value of the decodePhoto to `Bitmap?` as `BitmapFactory.decodeByteArray` can return null if it couldn't decode the photo.

app/src/main/java/com/epfl/Utils/drawyourpath/Utils.kt
```diff
@@ -33,23 +38,128 @@ object Utils {
...
     /**
-     * Encodes a bitmap image.
-     * @return A base64 string of the image.
+     * Decodes the photo from byte array to bitmap format
+     * @param photo photo encoded
+     * @return the photo in bitmap format
      */
-    fun encodePhoto(photo: Bitmap, quality: Int = 50): String {
+    fun decodePhoto(photo: ByteArray): Bitmap? {
+        return BitmapFactory.decodeByteArray(photo, 0, photo.size)
```
clementcharmillot, line 65: While testing, this function seems to throw sometimes when the conversion fails. Not sure if we want to handle that here. I let you decide.

clementcharmillot (approved):
Thanks for the changes. Just a small remark and we can merge, good work.

app/src/main/java/com/epfl/Utils/drawyourpath/Utils.kt
HoLuDyEr: This function can throw IndexOutOfBound if the size is bigger than the bytearray (which should never happen here) or IllegalArgumentException but only if some options are specified, here they are not specified so should also never happen.

PR #150, opened 2023-05-04
Code cleanup

CristianSafta:
Did some code cleanup on the FriendsFragment.


app/src/main/java/com/epfl/drawyourpath/mainpage/fragments/FriendsFragment.kt
```diff
@@ -46,18 +44,31 @@ class FriendsFragment(private val database: Database) : Fragment(R.layout.fragme
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
 
-        val database: Database = this.database
```
Gafragni, line 49: Just a question, is this not used in the handleUsernameSearch ?

app/src/main/java/com/epfl/drawyourpath/mainpage/fragments/FriendsFragment.kt
```diff
@@ -70,85 +81,88 @@ class FriendsFragment(private val database: Database) : Fragment(R.layout.fragme
...
         val userAccountFuture = database.getUserData(currentUser.getUid())
 
         userAccountFuture.thenApply { userdata ->
-
-            // Initialize the ViewModel with the userModel
             val factory = FriendsViewModelFactory(userdata.userId!!, this.database)
             viewModel = ViewModelProvider(this, factory)[FriendsViewModel::class.java]
 
-            // Update the adapter with the fetched data
-            recyclerView.adapter = friendsListAdapter
```
Gafragni, line 84: Just to make sure I understand, we do not update the adapter as it was already assigned on line 59 (65), right ?

Gafragni (approved):
Overall good refactoring where it was most needed !

Just commented two questions to make sure I understood the code correctly (as I am new to this part of the code)

app/src/main/java/com/epfl/drawyourpath/mainpage/fragments/FriendsFragment.kt
CristianSafta: correct

app/src/main/java/com/epfl/drawyourpath/mainpage/fragments/FriendsFragment.kt
CristianSafta: It is but I added it as an argument to the constructor so it simplifies the testing process(aka when you have to use mock database instead of the firebase)
class FriendsFragment(private val database: Database) : Fragment(R.layout.fragment_friends).
I don't remember specifically what that line was for but when searching the database from the argument above is used.

PR #153, opened 2023-05-04
fixed add/remove firebase function + friend fragment

HugoHof:
This PR fixed the problem that break the friendlist displaying.
Resolve add/remove friend function + correct friend fragment


Gafragni (approved):
Good catch on the friends display !
LGTM

PR #154, opened 2023-05-04
removed duplicate string resource

CristianSafta:
Bug fix for duplicate string resources


HugoHof (approved):
Thank you the bug fix

PR #168, opened 2023-05-05
167 fix set username and get goals from firebase

HugoHof:
This PR pic the set username function (the username was not correctly updated in the profile of the user) and getData function that break the retrieve of the runs and daily goals.


clementcharmillot (approved):
Thanks for the quick fix. We can merge

PR #182, opened 2023-05-09
171 bitmap from ink

### LINKED ISSUE #171: Create a bitmap from an ink object

clementcharmillot:
*Main Task*: 
- Adds a method to normalize the list of points so it fits in the unity square.
- Adds methods to convert a list of coordinates and a stroke to a bitmap image.

*Note*: Coverage for bitmap conversion will increase when it will be shown in the UI.

*Other Minor (not in their own PR because it's really small)*: 
- Changes Auto-format to validate code only without pushing it, the task fails if code doesn't respect coding format. Reason is that codeclimate fails with code pushed by the bot anyways. Changed the action as required for merging. Run `./format.bat` in root directory to auto format before committing.
- Add some tests for other parts.
- Refactor `getStringDuration` and `getStringTimeStartEnd`.
- Rename `PathDrawingCurrentPerformanceFragmentTest` to `PathDrawCurPerfFragTest` as it exceeds the maximum CL size.

closes #171


app/src/main/java/com/epfl/drawyourpath/utils/Utils.kt
```diff
@@ -264,4 +247,128 @@ object Utils {
...
+    /**
+     * Gets the smallest coordinates independently.
+     * When we want to fit the path in a square, it represents the topleft corner of the square.
+     * @return The smallest coordinate for the stroke.
+     */
+    fun getSmallestPoint(stroke: Stroke): Point =
+        stroke.points.fold(Pair(Float.MAX_VALUE, Float.MAX_VALUE)) {
+                acc, point ->
+            Pair(min(acc.first, point.x), min(acc.second, point.y))
+        }.let { Point.create(it.first, it.second) }
```
HoLuDyEr, line 272: These two functions are quite similar even if one uses the other. One alternative I can think of is : 
```kotlin
fun getSmallestPoint(strokes: List<Stroke>): Point =
        strokes.map { it.points }.flatten().fold(Pair(Float.MAX_VALUE, Float.MAX_VALUE)) { acc, point ->
            Pair(min(acc.first, point.x), min(acc.second, point.y))
        }.let { Point.create(it.first, it.second) }
```
and the other
```kotlin
fun getSmallestPoint(stroke: Stroke): Point = getSmallestPoint(listOf(stroke))
```

HoLuDyEr (changes requested):
Overall it is a good PR. The code is clean, easy to understand and well documented.
I only have a small remark for some code duplication.

clementcharmillot:
Thanks for the review, I've addressed both remarks. Let me know if there's anything more.

HoLuDyEr (approved):
Thank you for the changes.
You can merge now.

PR #186, opened 2023-05-14
180 reformat run path more stats

### LINKED ISSUE #180: Reaformat Run and Path to add duration and more precise statistics on the run

HugoHof:
This PR contains:
- modification of the path such that a path is now composed of a List<List<Latng>> instead of List<LatLng. Suchat that the user can draw path with multiple section. A new section is begin each time the user made a pause/resume in his drawing. Add and modify the methods and tests to be adapted to this new structure.
- modification of the run by adding a duration field. Now the duration is not directly relative to start time and end time. The duration is a separate parameter. 
- adapt the drawing model to the new structure of the run and path.
- add methods in Run class to give moe precise statistic of a run such that time per kilometer frame, average speed per kilometer frame, distance per section frame, time per section frame and average speed per section frame. 

Reamark: lots of tests and code in cache/database relative to run are commented because will be implemented in other task of this sprint. #174, #175, #176

Thank you in advance for your review.
Close #180


HoLuDyEr (commented):
This is a good PR, the code is clean and well documented. There are only some small remarks.
I will update the cache on another branch, then do a PR to merge on this branch before merging this branch on the main branch to avoid problems with the cache.

PR #190, opened 2023-05-15
175 Store ML Result in Firebase + Update Path Storage

### LINKED ISSUE #175: Save form recognized + score given by ML model in Firebase

### LINKED ISSUE #176: Adapt the way to store runs in Database (with multiple paths)

damiengenoud:
This PR modifies the class Run, so that it contains a shape (predicted by the ML model) and the score that was given to the path, relative to the shape. 

It also adapts the way Runs are stored in the history now that a Path can have multiple sections.

All the commented tests have been adapted to pass with the modified Run class. 

closes #175, closes #176


app/src/main/java/com/epfl/drawyourpath/database/FirebaseDatabaseUtils.kt
```diff
@@ -80,27 +82,36 @@ object FirebaseDatabaseUtils {
...
+                val lat = getNumber(point.child("latitude"))?.toDouble()
+                val lon = getNumber(point.child("longitude"))?.toDouble()
+                if (lat != null && lon != null) {
+                    LatLng(lat, lon)
+                } else {
+                    Log.w(this::class.java.name, "A coordinate was badly formatted.")
+                    null
+                }
             }
         } ?: emptyList()
```
clementcharmillot, line 96: Map returns an empty list so I don't think we need to filter `null` values.

```suggestion
        val sections = data?.child("path")?.child("points")?.children?.map { section ->
            section.children.mapNotNull { point ->
                val lat = getNumber(point.child("latitude"))?.toDouble()
                val lon = getNumber(point.child("longitude"))?.toDouble()
                if (lat != null && lon != null) {
                    LatLng(lat, lon)
                } else {
                    Log.w(this::class.java.name, "A coordinate was badly formatted.")
                    null
                }
            }
        } ?: emptyList()
```

clementcharmillot (approved):
Thanks for the PR, the code is clean and there isn't much to say. Two small minors remarks and we can merge. Nice work.

damiengenoud:
@clementcharmillot 
Thanks for the review, I made the changes you requested.

PR #194, opened 2023-05-18
170 smooth path

### LINKED ISSUE #170: Smoothen points with Ramer-Douglas-Peucker algorithm

Gafragni:
# The Ramer-Douglas-Peucker algorithm for paths
## The public function interfaces
This PR implements an algorithm to reduce the number of points on a path.
The `reducePath` function takes two arguments, the path and the maximal error allowed. The maxError argument is simply the percentage of error allowed on each point relative to the total path length. 

For example if we have a path of length 1234 meters and a maxError of 0.01 (default) then the algorithm will find a similar path with number of points <= to the given path, by removing points that are less than 12.34 meters away from the simplified path.

The `reduceSection` function is similar, but only reduces a single section and takes a different argument epsilon, which is the maximal distance error the algorithm is allowed to make per point.
## The distance 
The most challenging part is to describe the distance between points and a segment given the latitude and longitude coordinates. Ideally we would like to compute it on the spheroid (ellipsoid of revolution) but for this implementation we will use a spherical approximation in some steps.

In order to make it understandable and mainainable the distance is computed in substeps :
- Translate the latitude and longitude (rotation of the sphere) of the points to put the starting point of the segment in (0, 0) for latitude and longitude, or equivalently (1, 0, 0) in xyz coordinates.
- Find the angle the end point of the segment makes in the perpendicular plane (yz plane).
- Rotate around the x axis to put the segment on the equator (starting point still at the "origin").
- Now the latitude of the points can be used to calculate the distance to the line (more precisely circle) going through the segment.
- Use the longitude to determine what is the final distance of the points to the segment (it can be the distance to the starting point, the distance to the end point or the distance to the circle going through the segment).

The advantage of this approach is that in the end we obtain 3 possible points of reference, so we can use the more precise WGS84 (spheroid) distance calculation to get the final distance.

closes #170


app/src/main/java/com/epfl/drawyourpath/utils/Utils.kt
```diff
@@ -392,4 +394,118 @@ object Utils {
...
+            return Path(path.getPoints())
+        }
+
+        val epsilon = maxError * path.getDistance().toFloat()
+        for (section in path.getPoints()) {
+            reducedPoints.add(reduceSection(section, epsilon))
+        }
+        reducedPoints.removeFirst()
+        return Path(reducedPoints.toList())
+    }
```
damiengenoud, line 416: I think this function can be simplified a bit by returning `path` when the algorithm is not used, and by infering the type of `reducedPoint` instead of adding an empty list.
```suggestion
    fun reducePath(path: Path, maxError: Float = 0.01F): Path {
        val reducedPoints = mutableListOf<List<LatLng>>()
        if (path.size() <= 2) {
            return path
        }

        val epsilon = maxError * path.getDistance().toFloat()
        for (section in path.getPoints()) {
            reducedPoints.add(reduceSection(section, epsilon))
        }
        return Path(reducedPoints.toList())
    }
```

app/src/main/java/com/epfl/drawyourpath/utils/Utils.kt
```diff
@@ -392,4 +394,118 @@ object Utils {
...
+     * @param section The section we want to reduce
+     * @param epsilon The min distance to keep the point on the section
+     * @return A similar segment with the least useful points removed.
+     */
+    fun reduceSection(section: List<LatLng>, epsilon: Float): List<LatLng> {
+        val reducedPointList = mutableListOf<LatLng>()
+        if (section.size <= 2) {
+            reducedPointList.addAll(section)
+            return reducedPointList.toList()
+        }
```
damiengenoud, line 429: Same here, you can probably just return `section`
```suggestion
        if (section.size <= 2) {
            return section
        }
```

app/src/main/java/com/epfl/drawyourpath/utils/Utils.kt
```diff
@@ -392,4 +394,118 @@ object Utils {
...
+        if (distMax <= epsilon) {
+            reducedPointList.add(section.first())
+            reducedPointList.add(section.last())
+        } else {
+            val firstHalf = reduceSection(section.subList(0, indexMax + 1), epsilon)
+            val secondHalf = reduceSection(section.subList(indexMax, section.size), epsilon)
+            reducedPointList.addAll(firstHalf.subList(0, firstHalf.size - 1))
+            reducedPointList.addAll(secondHalf)
+        }
+        return reducedPointList.toList()
```
damiengenoud, line 444: Clean recursive algorithm ! Do you think there is a risk that the algorith has to do a very deep recursion that could cause memory issues ?

app/src/main/java/com/epfl/drawyourpath/utils/Utils.kt
```diff
@@ -392,4 +394,118 @@ object Utils {
...
+        var distances = mutableListOf<Float>()
+        for (point in points) {
+            // Translate to put the start at the origin and rotate to put the segment on the equator
+            val newPoint = rotateXAxis(
+                point.latitude - start.latitude,
+                point.longitude - start.longitude,
+                -anglePlane,
+            )
+            // Compute the distance to the segment case by case
+            var results = FloatArray(3)
```
damiengenoud, line 475: `distance` and `result` should be val, not var.

app/src/main/java/com/epfl/drawyourpath/utils/Utils.kt
```diff
@@ -392,4 +394,118 @@ object Utils {
...
+        val x = cos(latRadian) * cos(lonRadian)
+        val y = cos(latRadian) * sin(lonRadian)
+        val z = sin(latRadian)
+        // Apply rotation along the X axis (equivalent to a 2D rotation on y and z)
+        val newX = x
+        val newY = cos(angle) * y - sin(angle) * z
+        val newZ = sin(angle) * y + cos(angle) * z
+        // Transform back to latitude and longitude
+        val newLat = asin(newZ)
+        val newLon = atan2(newY, newX)
```
damiengenoud, line 508: The variable `newX` is basically useless, you can just keep `x`
```suggestion
        val newY = cos(angle) * y - sin(angle) * z
        val newZ = sin(angle) * y + cos(angle) * z
        // Transform back to latitude and longitude
        val newLat = asin(newZ)
        val newLon = atan2(newY, x)
```

damiengenoud (changes requested):
Nice PR, well commented (that was necessary with all the math) and documented! I did not check on the math formula myself, but the implementation seems fine. Just a few remarks or questions and then we can merge.

app/src/main/java/com/epfl/drawyourpath/utils/Utils.kt
Gafragni: Right I forgot that we do not need to instanciate the list directly.
For the path I wanted to make a copy but that is not necessary.

Your suggestion works perfectly then

app/src/main/java/com/epfl/drawyourpath/utils/Utils.kt
Gafragni: Same here, I was making a copy but it is not required

app/src/main/java/com/epfl/drawyourpath/utils/Utils.kt
Gafragni: Right, the content of the list is modified, not the list themselves.

app/src/main/java/com/epfl/drawyourpath/utils/Utils.kt
Gafragni: Yes as the rotation is only around the X axis the x coordinate does not change.
I'll remove the useless variable and put a comment instead

app/src/main/java/com/epfl/drawyourpath/utils/Utils.kt
Gafragni: It is a neat algorithm ! For the performance the algorithm is $\Theta(n^2)$ and in the ideal case (when the recursion splits the work equally into two subtasks) it is $O(n*log(n))$, where n is the number of points in the original path.

The depth of the recursion should not be an issue : 
If we call m the number of segment in the reduced path then the depth of the recursion is
- log(m) in the ideal case
- m in the worst case 

So if we have a very long path that cannot be reduced well (like a very long zigzag) and that we mess with the epsilon parameter (the `reducePath` function should compensate with the length) then yes, but if you use the functions normally there should not be any issues whatsoever.

damiengenoud: Alright thanks for the clear answer. If it's not too risky then it's fine for me!

damiengenoud:
Thanks for the changes, you can merge! Good job!

damiengenoud (approved):
Thanks for the changes, you can merge! Good job!

PR #214, opened 2023-05-21
Fixed the bug where reloading the friends fragment causes the app to crash sometimes.

CristianSafta:
Fixed a bug where clicking fast on the friends fragment would cause the app to crash.


HugoHof (approved):
Thank you for the bug fix

PR #229, opened 2023-05-25
204 more statistics

### LINKED ISSUE #204: Add statistics on the evolution of runs over time

Gafragni:
This PR implements more functions on statistics and links the data to the functions and the UI.

The statsFragment may be refactored but the rest is clean and maintainable.

closes #204


app/src/main/java/com/epfl/drawyourpath/userProfile/cache/UserModelCached.kt
```diff
@@ -66,6 +66,11 @@ class UserModelCached(application: Application) : AndroidViewModel(application)
         .map { entity -> entity?.let { UserProfile(it) } ?: UserProfile(UserEntity("empty user")) }
 
     // dailyGoal
+    private val dailyGoals: LiveData<List<DailyGoal>> = user.switchMap { user ->
+        dailyGoalCache.getDailyGoalById(user.userId).map { dailyGoalEntities ->
+            dailyGoalEntities.map { DailyGoal(it) }
+        }
+    }
```
HoLuDyEr, line 73: this could be refactored as 
```kotlin
private val dailyGoals: LiveData<List<DailyGoal>> = _currentUserID.switchMap {
        dailyGoalCache.getDailyGoalById(it)
    }.map { entities -> entities.map { DailyGoal(it) } }
```
this only needs the userId so no need for the user

HoLuDyEr (changes requested):
Thank you for this PR. The code is clean, well tested and well documented. There is only a small remark.

HoLuDyEr (approved):
Thank you for the changes, you can merge now.

PR #232, opened 2023-05-26
add empty posts at the end

HoLuDyEr:
This PR fixes an UI issue where the add post button was on top of the votes of the last, now we can scroll more to vote on it.

This PR also fixes a bug with the graph library for the stats that now shows default values (0.0) for day/month where the user has no dailygoal.


HugoHof (approved):
Thank you for the bug correction and re implement the test of the trophies in a new file such that they compile.

